Issue #1070, opened 2026-02-05
[PR2c] Blob transfer state machine + resource bounds


fahertym:
## Summary
Implement blob transfer state machine with resource bounds and cleanup.

## Depends On
- #1069 (PR2b: ReplayGuard wiring)
- #1066 (PR1a: BlobStore impl)

## State Machine
```
Requested → Receiving(chunks_seen) → Assembled → Verified → Stored
                                           ↓
                                    Failed (terminal)
```

## Tasks
- [ ] Create `icn-gossip/src/handlers/blob_sync.rs`
- [ ] Implement state machine with states above
- [ ] Temp dir for partial blobs: `data/blobs_partial/<request_id>/...`
- [ ] Enforce resource bounds:
  - Max in-flight requests per peer = 8 → `ErrCode::Busy`
  - Max concurrent reassemblies = 16 global → `ErrCode::Busy`
  - Expiry enforcement → `ErrCode::Expired`
  - Wrong source DID → `ErrCode::Forbidden`
  - Missing request_id → `ErrCode::InvalidRequest`
- [ ] Chunk hash verification before write
- [ ] Final hash verification before store
- [ ] Cleanup job (interval timer) removes expired partial dirs

## Files
- `icn-gossip/src/handlers/blob_sync.rs` (new)
- `icn-gossip/src/handlers/mod.rs`

## Acceptance Tests
- `reject_unsolicited_blob_transfer`
- `reject_blob_transfer_wrong_request_id`
- `reject_blob_transfer_after_expiry`
- `blob_transfer_reassembly_and_verification`
- `max_inflight_requests_enforced`
- `max_concurrent_reassemblies_enforced`
- `state_machine_cleanup_on_expiry`

## DoD
Node only sends transfer chunks TO requester. No broadcast.

## Failure Modes
- Partial chunks never complete → expiry cleanup
- Disk full → `ErrCode::Busy` + log disk error

## LOC Target
≤400 (may need split)

## Rollback
Delete `data/blobs_partial/` temp directory.

## Invariant Impact
- [x] Adversarial-by-default: Unsolicited transfers rejected, resource bounds prevent DoS
- [x] No panics: All error paths use ErrCode

PR #1107, opened 2026-02-06
feat(gossip): blob transfer state machine (#1070)

fahertym:
## Summary

Implements the blob transfer session lifecycle (PR2c) that turns signed chunks into a verified artifact with an `ArtifactReceipt` proof. Builds on PR2a (#1104, message types) and PR2b (#1106, nonce replay guard).

### What's included

- **ArtifactReceipt proof type** (`icn-kernel-api/src/proofs.rs`): Self-authenticating receipt with blake3 binding hash, tamper detection via `verify_binding()`, signed by requester
- **TransferSessionManager** (`icn-gossip/src/handlers/blob_transfer_state.rs`): Monotonic state machine (Requested → Receiving → Assembling → Verified → Stored / Failed) with resource bounds (8 max/peer, 16 concurrent reassemblies, 64KB chunks, 10MB max blob)
- **Handler wiring** (`icn-gossip/src/handlers/blob_transfer.rs`): Provider serves chunks from BlobService; requester feeds into state machine with two-phase assembly (lock→mark→unlock→IO→lock→finalize)

### Key design decisions

- **Two-phase assembly**: Prevents stalling chunk handling during disk IO + hash verification
- **Self-authenticating receipts**: `receipt_hash` computed at construction via blake3, `verify_binding()` recomputes for tamper detection
- **All rejections use stable ErrCode**: No new error codes added (InvalidRequest, Forbidden, Expired, Oversize, Busy)
- **Per-peer resource limits**: MAX_REQUESTS_PER_PEER=8, MAX_CONCURRENT_REASSEMBLIES=16

## Commits

1. `feat(kernel-api): add ArtifactReceipt proof type` — 8 tests
2. `feat(gossip): add blob transfer state machine` — 10 tests
3. `feat(gossip): wire transfer state machine into blob handlers` — 2 integration tests

## Invariants checklist

- [x] **Adversarial-by-default**: Per-peer limits, nonce replay guard, chunk hash verification, blob hash verification on assembly, expired session cleanup
- [x] **Determinism**: Receipt hash is deterministic (blake3 over canonical field ordering)
- [x] **Canonical encodings**: ContentHash = [u8; 32] in gossip, Hash = [u8; 32] in kernel-api
- [x] **No panics**: All error paths return Result with ErrCode, no unwrap/expect in protocol paths
- [x] **Kernel/app boundary**: ArtifactReceipt in kernel-api uses generic types (Hash, Did, Signature), no domain imports

## Test plan

- [x] `cargo test -p icn-kernel-api` — 148 tests pass (8 new proof tests)
- [x] `cargo test -p icn-gossip` — 211 tests pass (10 new state machine + 2 new integration)
- [x] `cargo clippy -p icn-kernel-api -p icn-gossip --all-targets -- -D warnings` — clean
- [x] `cargo fmt --check` — clean

## Risk

- `send_callback` visibility changed from private to `pub(crate)` — only accessible within the gossip crate
- GossipActor gains two new optional fields (blob_store, transfer_manager) — initialized as None, no behavioral change without explicit setup

### DIFF
--- a/.github/actions/install-build-tools/action.yml
+++ b/.github/actions/install-build-tools/action.yml
@@ -1,0 +1,19 @@
+name: 'Install Build Tools'
+description: 'Install clang and mold linker required by .cargo/config.toml'
+
+runs:
+  using: 'composite'
+  steps:
+    - name: Install clang and mold linker
+      shell: bash
+      run: |
+        PACKAGES=""
+        if ! command -v clang &> /dev/null; then
+          PACKAGES="$PACKAGES clang"
+        fi
+        if ! command -v mold &> /dev/null; then
+          PACKAGES="$PACKAGES mold"
+        fi
+        if [ -n "$PACKAGES" ]; then
+          sudo apt-get update -qq && sudo apt-get install -y -qq $PACKAGES
+        fi
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -38,6 +38,9 @@ jobs:
       - name: Free disk space
         uses: ./.github/actions/free-disk-space
 
+      - name: Install build tools (clang + mold linker)
+        uses: ./.github/actions/install-build-tools
+
       - uses: dtolnay/rust-toolchain@stable
 
       - uses: Swatinem/rust-cache@v2
@@ -75,6 +78,16 @@ jobs:
 
       - name: Free disk space
         uses: ./baseline/.github/actions/free-disk-space
+
+      - name: Install build tools (clang + mold linker)
+        shell: bash
+        run: |
+          PACKAGES=""
+          if ! command -v clang &> /dev/null; then PACKAGES="$PACKAGES clang"; fi
+          if ! command -v mold &> /dev/null; then PACKAGES="$PACKAGES mold"; fi
+          if [ -n "$PACKAGES" ]; then
+            sudo apt-get update -qq && sudo apt-get install -y -qq $PACKAGES
+          fi
 
       - uses: dtolnay/rust-toolchain@stable
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -85,6 +85,9 @@ jobs:
     steps:
       - uses: actions/checkout@v6
 
+      - name: Install build tools
+        uses: ./.github/actions/install-build-tools
+
       - uses: dtolnay/rust-toolchain@stable
 
       - name: Check kernel crates for forbidden deps (bash version)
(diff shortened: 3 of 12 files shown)


fahertym:
**Security Audit failure**: `RUSTSEC-2026-0009` (`time` v0.3.46, severity 6.8 medium).

Affects `main` too — not caused by PR2c code.

Fixed in a392ccbf: lockfile bump `time` 0.3.46 → 0.3.47, `time-macros` 0.2.26 → 0.2.27 via `cargo update time@0.3.46`. Patch-level transitive dep update only.

fahertym (commented):
## Code Review (Claude)

### 1. `active_reassemblies` counter leaks on error paths
**File**: `icn-gossip/src/handlers/blob_transfer_state.rs` — `complete_assembly()`
**Confidence**: 90/100

If `complete_assembly()` returns early on error (e.g., hash mismatch, IO failure), the `active_reassemblies` atomic counter is decremented via the `_guard` drop, but the session's `per_peer_counts` entry for that peer is never decremented because `cleanup_session()` is only called on the success path. Over time, peers that trigger assembly failures will hit the `MAX_SESSIONS_PER_PEER` (8) limit and be unable to start new transfers.

**Suggestion**: Ensure `per_peer_counts` is decremented in all exit paths of `complete_assembly()`, not just the success path. A drop guard pattern (similar to `_guard` for `active_reassemblies`) would be the most robust approach.

---

### 2. `SystemTime::now()` used instead of `icn_time::current_timestamp_secs()`
**File**: `icn-gossip/src/handlers/blob_transfer_state.rs`
**Confidence**: 85/100

The blob transfer state machine uses `SystemTime::now()` directly for session timeouts. The project has `icn_time::current_timestamp_secs()` for consistent time handling. Using raw `SystemTime` bypasses any time abstraction the project provides (e.g., for testing with mock clocks).

This was also flagged in a previous review of this PR.

---

### 3. Stale "stub handlers" documentation (3 instances)
**File**: `icn-gossip/src/handlers/blob_transfer.rs`
**Confidence**: 92/100

Module-level and function-level doc comments still refer to "stub handlers" or "placeholder" implementations, but the code now contains full working implementations:
- Module doc: claims "stub handlers" but contains complete chunk-serving and assembly logic
- `handle_blob_request()`: doc says "stub" but implements full provider response with chunked transfer
- `handle_blob_transfer_chunk()`: doc says "stub" but implements full reassembly and verification

These stale docs are misleading for anyone reading the code.

---

🤖 Generated with [Claude Code](https://claude.ai/code)

Issue #1310, opened 2026-03-01
feat(governance): ExecutionReceiptGate — enforce governance→execution linkage


fahertym:
Accepted governance proposals that authorize resource allocation currently have no enforcement that the allocation actually occurs. Add ExecutionReceiptGate middleware: resource-allocation journal entries are blocked until linked to a valid GovernanceDecisionReceipt. Track 'open' (voted, not yet executed) vs 'closed' (executed, receipt issued) governance decisions. Closes Invariant 7: execution receipts close the governance loop. Phase B of [EPIC] #1302.

fahertym:
Completed via PR #1327 (merged commit 977d8dab). ExecutionReceiptGate enforces Invariant 7 — governance→execution linkage.

PR #1327, opened 2026-03-06
feat(governance): ExecutionReceiptGate — Invariant 7 enforcement

fahertym:
## Summary

- Adds `icn_governance::gate::check_execution_gate()` — pure, stateless gate enforcing **Invariant 7**: resource-allocation journal entries must be backed by a `GovernanceDecisionReceipt` with outcome `Accepted` and a verified `decision_hash`
- `GateError::OutcomeNotAccepted` for `Rejected`/`NoQuorum` receipts
- `GateError::DecisionHashInvalid` for receipts whose fields were mutated post-construction (tampering/corruption detection via `receipt.verify()`)
- Re-exported at crate root: `icn_governance::{check_execution_gate, GateError}`

> **Scope:** This PR adds the gate primitive; a follow-up PR will wire it into execution entrypoints. The current allocation path consumes a raw `decision_hash: [u8; 32]`, not a `GovernanceDecisionReceipt`, so enforcement requires a signature change at the execution boundary. See the inline comment for the planned wiring approach.

## Design notes

Gate lives in `icn-governance` (not `icn-ledger`) to avoid a circular dependency: `icn-governance` already depends on `icn-ledger`, and `GovernanceDecisionReceipt` is defined in `icn-governance`. The double-execution guard (preventing re-use of a decision hash) is handled separately by `ExecutionRecord::is_terminal()` in the execution store — that's a stateful concern; this gate is intentionally stateless.

## Test plan

- [x] `test_gate_passes_accepted` — happy path
- [x] `test_gate_rejects_rejected_outcome` — outcome check
- [x] `test_gate_rejects_no_quorum` — outcome check
- [x] `test_gate_rejects_tampered_hash` — hash integrity
- [x] `test_error_display` — human-readable error messages
- [x] `cargo clippy -p icn-governance -- -D warnings` — clean
- [x] `cargo fmt --all` — formatted

Sprint 14 · task t16

### DIFF
--- a/icn/apps/governance/src/actor.rs
+++ b/icn/apps/governance/src/actor.rs
@@ -11,19 +11,19 @@ use std::str::FromStr;
 use std::sync::Arc;
 use std::time::{Duration, Instant};
 use tokio::sync::{mpsc, RwLock};
-use tracing::{debug, info, warn};
+use tracing::{debug, error, info, warn};
 
 use crate::state_store::{GovernanceStateStore, SledGovernanceStateStore};
 use icn_gossip::GossipActor;
 use icn_identity::Did;
 
 use icn_governance::{
-    DecisionOutcome, Delegation, DelegationId, GovernanceConfig, GovernanceDomain,
-    GovernanceDomainId, GovernanceMessage, GovernanceParams, GovernanceProfile,
-    GovernanceProfileId, MembershipAction, MembershipConfig, MembershipResolver, MembershipSource,
-    PaginatedResult, ParameterChange, Proposal, ProposalId, ProposalOutcome, ProposalPayload,
-    ProposalScope, ProposalState, ProtocolParameter, ProtocolParameterStore, TallySnapshot,
-    Timestamp, Vote, VoteChoice, VoteTally,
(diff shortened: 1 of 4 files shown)


fahertym:
## Call-site coverage note (for reviewers)

This PR introduces the gate *primitive*. As of this diff, `check_execution_gate` has **zero production call sites** — the existing `create_budget_allocation` in `icn-ledger/src/allocations.rs` accepts a raw `Hash`, not a `GovernanceDecisionReceipt`, so it can't call the gate without a signature change.

The correct enforcement point is wherever a caller holds both a `GovernanceDecisionReceipt` and is about to commit an allocation. That wiring requires either:
- changing `create_budget_allocation` to accept `&GovernanceDecisionReceipt` and call the gate internally, or
- adding a gated wrapper that callers use instead

Both options touch the icn-ledger public API and should land in a follow-on task (tracked under issue #1310 notes). This PR intentionally stops at the primitive to keep the diff reviewable in isolation.

If you want the enforcement wired in this PR, flag it and I'll extend scope. Otherwise, the follow-on is well-defined.

PR #1434, opened 2026-03-25
feat(provenance): governance→economics chain binding + JWT auth proof in direct-member entries

fahertym:
## Summary

Three provenance-hardening commits: INV-1 (direct-member auth evidence), INV-2 (allocation completeness), INV-5 (chain walkability), INV-6 (duplicate vote rejection), and a correction of a pre-existing `chain_complete` logic defect.

Addresses sprint task s26-t6.

## Problem

1. **INV-2:** Accepted economic proposals (`Budget`, `Treasury`, `Allocation`, `SurplusAllocation`) created a `GovernanceDecisionReceipt` but no downstream `AllocationReceipt`. The governance→economics chain was incomplete and unqueryable.

2. **INV-5:** No HTTP endpoint existed to query the provenance chain. Inspection required direct node access.

3. **`chain_complete` defect:** The chain completeness flag reduced to `has_governance` under all conditions due to the tautological expression `has_governance && (is_economic == has_allocations)`. An accepted Budget proposal with a missing allocation receipt would report `chain_complete: true` — a silent false guarantee.

4. **INV-1:** Every `DirectMember` journal entry stored the static string `"jwt-authenticated"` regardless of actual caller context — including escrow release and recurring settlements which have no JWT. No inspectable authorization evidence, no link to any specific request.

## Changes

**`apps/governance/`**
- `GovernanceReceiptBackend` trait: adds `put_allocation`, `get_governance_by_decision`, `list_allocations_by_decision`
- `GovernanceManager::create_allocation_receipt`: creates `AllocationReceipt` for economic payloads on `ProofOutcome::Accepted`; wired into `close_proposal`
- `GovernanceManager::get_chain`: returns `ProvenanceChain { governance_receipt, allocations, chain_complete }` with corrected completeness logic
- `GET /gov/proposals/{id}/chain` registered
- 5 invariant regression tests (INV-2×2, INV-5×2, INV-6)

**`crates/icn-gateway/`**
- `LedgerManager::create_settlement` takes `auth_proof: Option<String>`
- HTTP handler (`POST /ledger/{coop_id}/settle`) extracts JWT third segment from `Authorization: Bearer <h>.<p>.<sig>`, passes `Some("jwt-sig:<sig>")`
- Missing/malformed Authorization header falls through to `None` → stored `"system-executed"` (no false proof claim)
- Escrow release and recurring settlement callers pass `None`
- `GET /health/full` composite health endpoint with per-subsystem probes and latency
- 2 INV-1 provenance-readback regression tests via `get_history()` at `ProvenanceRef::DirectMember.signature`

## Verification

- `cargo check -p icn-gateway`: clean
- `cargo fmt --all --check`: clean
- `cargo clippy -p icn-gateway --all-targets -- -D warnings`: clean
- `cargo test -p icn-gateway --lib`: 458/458 pass
- INV-1 tests verify storage truth via `get_history()` readback — not just call-site threading

## Semantics

The stored `"jwt-sig:<segment>"` is the HMAC-SHA256 fingerprint of the JWT that authorized the HTTP action. It ties the journal entry to a specific, verified auth token. It is **not** a public-key signature over entry content. Nodes holding `jwt_secret` can verify by recomputing `HMAC-SHA256(secret, "{header}.{payload}")`.

System-executed paths store `"system-executed"` — an honest marker that no per-request proof exists.

## Limitations

- HMAC verification requires `jwt_secret`. Federation replicas and external auditors without the secret cannot verify stored fingerprints.
- The `signature` field name in `ProvenanceRef::DirectMember` implies Ed25519; the stored value is an auth-token fingerprint, not a signature over entry content.
- No HTTP-level actix integration test covering the full extraction→storage path; proof is at the manager/storage boundary.
- Missing/malformed Authorization header degrades silently to `"system-executed"`. Indistinguishable from a system-executed entry at the storage layer.

## Follow-ups

- HTTP-level actix integration test: verify `api/ledger.rs` correctly extracts and threads the JWT third segment end-to-end
- Asymmetric per-entry signing over entry content using the node's Ed25519 key — publicly verifiable without secret sharing
- Disambiguate `ProvenanceRef::DirectMember.signature` field name from Ed25519 semantics

### DIFF
--- a/icn/apps/governance/src/http/configure.rs
+++ b/icn/apps/governance/src/http/configure.rs
@@ -96,6 +96,10 @@ where
             web::resource("/proposals/{proposal_id}/proof")
                 .route(web::get().to(handlers::get_proof::<E>)),
         )
+        .service(
+            web::resource("/proposals/{proposal_id}/chain")
+                .route(web::get().to(handlers::get_chain::<E>)),
+        )
         // ── Discussion endpoints ─────────────────────────────────────────
         .service(
             web::resource("/proposals/{proposal_id}/discussion")
--- a/icn/apps/governance/src/http/handlers.rs
+++ b/icn/apps/governance/src/http/handlers.rs
@@ -1118,6 +1118,31 @@ pub async fn get_vote_tally<E: GovernanceEventEmitter + Clone + 'static>(
     }))
 }
 
+/// GET /gov/proposals/{proposal_id}/chain — Get the full provenance chain for a proposal.
+///
+/// Returns the governance decision receipt and any linked allocation receipts,
+/// enabling independent verification of the governance→economics binding (INV-5).
+///
+/// Response includes:
+/// - `governance_receipt`: The decision receipt (present if proposal is closed)
+/// - `allocations`: AllocationReceipts created when the proposal was accepted
+/// - `chain_complete`: Whether the chain is complete for this proposal type
+///
+/// A `chain_complete: false` response means the system cannot answer
+/// "what economic effect did this decision have?" — a provenance gap.
+pub async fn get_chain<E: GovernanceEventEmitter + Clone + 'static>(
+    ctx: web::Data<GovernanceContext<E>>,
+    http_req: HttpRequest,
+    proposal_id: web::Path<String>,
+) -> Result<HttpResponse, ApiError> {
+    require_scope::<BasicClaims>(&http_req, "governance:read")?;
+
+    let id = ProposalId(proposal_id.into_inner());
+    let chain = ctx.manager.get_chain(&id).await.map_err(anyhow_to_api)?;
+
+    Ok(HttpResponse::Ok().json(chain))
+}
+
 /// GET /gov/proposals/{proposal_id}/proof — Get cryptographic proof of proposal outcome.
 pub async fn get_proof<E: GovernanceEventEmitter + Clone + 'static>(
     ctx: web::Data<GovernanceContext<E>>,
--- a/icn/apps/governance/src/manager.rs
+++ b/icn/apps/governance/src/manager.rs
@@ -19,9 +19,23 @@ use icn_governance::{
     DEFAULT_MAX_DELEGATION_DEPTH,
 };
 use icn_identity::Did;
+use icn_kernel_api::{AllocationReceipt, ScopeLevel, SettlementIntent};
 use std::collections::{HashMap, HashSet};
 use std::sync::{Arc, RwLock};
 use tracing::debug;
+
+/// Full provenance chain for a governance proposal (INV-5).
+///
+/// Links governance decision → allocation receipts for independent verification.
+#[derive(Debug, Clone, serde::Serialize)]
+pub struct ProvenanceChain {
+    /// Governance decision receipt (present if proposal was closed)
+    pub governance_receipt: Option<icn_governance::GovernanceDecisionReceipt>,
+    /// Allocation receipts linking decision to economic intents
+    pub allocations: Vec<AllocationReceipt>,
+    /// True if the chain is complete for this proposal type
+    pub chain_complete: bool,
+}
 
 // ============================================================================
 // Sled-Backed Action Item Store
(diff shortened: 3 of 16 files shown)


fahertym:
**HTTP extraction proof gap closed** (commit 4cde7237)

Two actix-level integration tests added to `src/api/ledger.rs` that prove the full JWT extraction→storage path:

- `test_inv1_http_settlement_stores_jwt_sig_in_provenance`: Sends `Authorization: Bearer hdr.payload.fakesig` through the real actix handler, reads back the journal via `get_history()`, asserts `DirectMember.signature == "jwt-sig:fakesig"`.
- `test_inv1_missing_bearer_sig_stores_system_executed_not_jwt_sig`: No Authorization header → asserts `"system-executed"`, not `"jwt-sig:*"`.

Auth bypass: `TokenClaims` injected via request extensions (middleware reads extensions; the `Authorization` header is read only by the INV-1 extraction path). The two mechanisms are orthogonal — this is the correct seam to probe.

Previously documented limitation in the PR description is now closed at the HTTP handler layer. The four INV-1 tests total: two at manager/storage boundary (`ledger_mgr.rs`), two at HTTP handler boundary (`api/ledger.rs`).

fahertym:
### Tranche 5 — #1437: http-auth-missing vs system-executed (`a994ac2c`)

**Problem:** Missing/unparseable `Authorization` on the HTTP path stored `"system-executed"` — indistinguishable from escrow/recurring paths that truly had no HTTP context.

**Mechanism chosen:** Resolve the fallback in the HTTP handler itself. The extraction chain now ends with `.unwrap_or_else(|| "http-auth-missing".to_string())`, and the handler always passes `Some(...)`. `create_settlement`'s signature is unchanged; non-HTTP callers still pass `None` → `"system-executed"`.

**New stored semantics:**
| Stored value | Path |
|---|---|
| `"jwt-sig:<seg>"` | HTTP path, Authorization header present and parseable |
| `"http-auth-missing"` | HTTP path, Authorization header absent or unparseable |
| `"system-executed"` | Non-HTTP system paths (escrow, recurring settlements) |

**Tests:** Renamed `test_inv1_missing_bearer_sig_stores_system_executed_not_jwt_sig` → `test_inv1_missing_bearer_sig_stores_http_auth_missing`, updated assertion. All 460 icn-gateway lib tests pass. fmt + clippy clean.

**Files:** `api/ledger.rs` (functional + test), `ledger_mgr.rs` (doc comment only).

Closes #1437

fahertym:
### Tranche 6 — #1435 enabling step: Ed25519 sign/verify substrate (`26a49366`)

**Audit result:** Path B (bounded enabling step). Full member-key signing is blocked by two key-access gaps:

1. **Member key unavailable:** HTTP settlement auth uses JWT (HMAC challenge-response). The member's Ed25519 private key never reaches the gateway server. The JWT proves identity; key material stays client-side.
2. **No node keypair in gateway:** `GatewayServer` holds only `jwt_secret: Vec<u8>`. `LedgerManager` has no key material. `IdentityManager` stores DID documents (public info only).

**What exists:** `JournalEntry.signature: Option<Signature>` is already defined for this purpose, excluded from content hash computation, and the pattern (sign `entry_hash.as_bytes()`, verify with DID-extracted public key) is already live in the witness signature code.

**What was added (`icn-ledger/src/entry.rs`):**
- `sign_entry(entry: &mut JournalEntry, keypair: &KeyPair) -> Result<()>` — signs the 32-byte content hash, stores 64-byte Ed25519 signature
- `verify_entry_signature(entry: &JournalEntry) -> Result<bool>` — verifies using only `entry.author` (DID = public key, no secret needed)

**Proof tests:** golden path, unsigned returns false, wrong keypair fails, content mutation breaks verification.

**Remaining gap:** Thread a signing capability into `LedgerManager::create_settlement`. Requires either (a) client-side pre-signing protocol or (b) node-level attestation with a gateway keypair that doesn't currently exist.

**Files:** `icn-ledger/src/entry.rs` only (substrate layer, no gateway changes).

PR #1554, opened 2026-04-17
feat(governance): materialize action items from accepted proposals (standalone path)

fahertym:
## Summary

The standalone/in-memory `close_proposal_inner` path in `GovernanceManager` now materializes `action_items_on_accept` specs into concrete `ActionItem`s when a proposal is accepted — matching the behavior already present in the actor-backed path (`actor.rs::materialize_action_items`).

## Changes

- **`apps/governance/src/manager.rs`**: Added materialization block in `close_proposal_inner` that fires when `outcome == Accepted && !proposal.action_items_on_accept.is_empty()`. Includes a dedup guard (checks if items already exist for this proposal ID before writing) to prevent double-materialization if both the actor and standalone paths ever run on the same proposal.
- **`apps/governance/src/http/handlers.rs`**: Added two tests:
  - `accepted_proposal_materializes_action_items`: proposal with 2 specs → 2 `ActionItem`s with correct `linked_proposal`, assignee, priority, due date, tags
  - `rejected_proposal_does_not_materialize_action_items`: rejected proposal → 0 items

## Motivation

Without this fix, any deployment using the HTTP-only/standalone manager path (not wired to the GovernanceActor) would silently drop obligation specs when proposals were accepted. The actor path was the only path that materialized action items. Since the `action_items_on_accept` field already exists on `Proposal` and is accepted via the HTTP API, this was a silent correctness gap.

This is the minimal governance-to-execution bridge: an accepted proposal can now produce tracked obligations as generic `ActionItem`s with full provenance, no CCL evaluation required.

## Testing

- [x] `accepted_proposal_materializes_action_items` — verifies count, assignee, priority, due_date, tags, linked_proposal
- [x] `rejected_proposal_does_not_materialize_action_items` — verifies zero items on rejection
- [x] 86/86 tests pass
- [x] `cargo fmt --check` clean
- [x] `cargo clippy` clean (0 warnings)

## Invariants

- [x] No panics introduced in protocol paths
- [x] Determinism preserved — materialization is deterministic given same specs
- [x] Canonical encodings unchanged
- [x] Trust gates not weakened
- [x] Kernel/app boundaries respected — changes confined to `apps/governance` (domain layer)

## Verification

```
cargo fmt --check -p icn-governance-actor  → exit 0
cargo clippy -p icn-governance-actor --all-targets -- -D warnings  → 0 warnings
cargo test -p icn-governance-actor --lib  → 86 passed; 0 failed
```

## Documentation

- [x] N/A — implementation of existing field semantics, no new API surface

### DIFF
--- a/icn/apps/governance/src/http/handlers.rs
+++ b/icn/apps/governance/src/http/handlers.rs
@@ -3966,4 +3966,253 @@ mod tests {
             "membership_resolver.resolve_members() must be invoked for TrustThreshold domains"
         );
     }
+
+    // ── Governance-to-execution bridge ──────────────────────────────────────
+
+    /// Proves: when a proposal with `action_items_on_accept` is accepted via
+    /// `close_proposal`, the specs are materialized as concrete `ActionItem`s
+    /// with correct provenance (`linked_proposal` set to the proposal ID).
+    ///
+    /// This tests the standalone/in-memory path (no actor). The actor path is
+    /// tested separately in `actor.rs`. Both paths must produce the same result.
+    #[tokio::test]
+    async fn accepted_proposal_materializes_action_items() {
+        // Seeds 1 and 2 are verified valid Ed25519 compressed points (used in
+        // hook_fires_with_correct_payload_on_acceptance).
+        let member_did = test_did(1);
+        let assignee_did = test_did(2);
+        let mgr = Arc::new(GovernanceManager::new());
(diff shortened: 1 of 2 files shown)


fahertym:
Superseded by #1555, which includes this bridge plus /gov/me/work filter support, postcard migration fix, and comprehensive tests. All content absorbed there.

PR #1559, opened 2026-04-17
feat(governance): CCL-backed milestone gates — storage, evaluation, HTTP API

fahertym:
## Summary

Adds an optional CCL-backed completion gate to milestones. When completion_gate is set, the CCL interpreter evaluates a declarative boolean expression against milestone context before accepting a Completed transition. A false result rejects completion with a clear error; no gate means existing behavior unchanged.

## Changes

- icn-governance/src/milestone_gate.rs (new): MilestoneGateContext, MilestoneGateError, evaluate_milestone_gate, parse_gate. Builds a synthetic single-require-rule CCL contract, evaluates it read-only against criteria_count / phase_index, maps PreconditionFailed to Blocked and other errors to EvalError.
- Milestone.completion_gate: Option<String>: JSON-encoded icn_ccl::Expr, stored as Option<String> for postcard/Sled compatibility. Uses serde(default) (not skip_serializing_if) to preserve the postcard discriminant byte.
- GovernanceManager::set_milestone_gate: public API to set/clear the gate on an existing milestone. Serializes Expr to JSON.
- GovernanceManager::update_milestone_status: parses the JSON gate on completion attempt, evaluates it, rejects if blocked. Non-completion transitions skip gate entirely.
- icn-governance Cargo.toml: adds icn-ccl dependency.
- icn-governance-actor Cargo.toml: adds icn-ccl dependency.

## Motivation

Milestone completion_criteria was passive text. This is the first truthful path where a milestone can be conditionally gated by declarative logic, following CCL evaluation model (deterministic, read-only, fuel-bounded) without hardcoding institution-specific rules.

## Gate context (first slice)

Variables available to the expression: criteria_count (Int) and phase_index (Int).

Explicitly deferred: action-item completion counts, trust scores, ledger balances, cross-milestone dependency checks.

## Testing

- [x] 8 unit tests in milestone_gate::tests: no gate, gate true, gate false, AND compound, literal true/false, unknown variable gives EvalError
- [x] 5 manager integration tests: no gate succeeds, gate true succeeds, gate false blocks (milestone unchanged), eval error surfaced, non-completion bypass
- [x] All 637 pre-existing tests continue to pass
- [x] cargo fmt --check clean
- [x] cargo clippy -- -D warnings clean

## Invariants

- [x] No panics in protocol paths
- [x] Determinism preserved (CCL evaluation is deterministic; fuel bounded at 1000)
- [x] Canonical encodings: postcard roundtrip preserved via Option<String>
- [x] Trust gates not weakened
- [x] Kernel/app boundaries respected

## Verification

cargo fmt --check -p icn-governance -p icn-governance-actor: clean
cargo clippy -p icn-governance -p icn-governance-actor -- -D warnings: clean
cargo test -p icn-governance -p icn-governance-actor: 637 tests, 0 failures

## Documentation

N/A - milestone_gate.rs is self-documenting with module-level doc and inline comments


fahertym:
Closing without merge. Superseded by the honest subset in #1561 (milestone preview + history + program summary + event log + PATCH /gov/programs/{id}/status).

Rationale — this branch has four blocking issues that together do not justify a fix-forward:

1. **P1 postcard storage-layout break.** `Milestone.completion_gate: Option<String>` was inserted mid-struct in `icn-governance::program`. Postcard uses positional encoding, so any persisted milestone predating this change is no longer decodable. `serde(default)` does not rescue postcard the way it rescues JSON — a new required position in the byte layout shifts all following fields.
2. **Non-atomic gate validation.** `GovernanceManager::update_milestone_status` evaluates the gate and persists the status via separate operations with no sled transaction bracketing them. A gate that passes at evaluation-time can be undermined by a concurrent `set_milestone_gate` write; a gate that persists is not guaranteed to have been enforced.
3. **Gate-failure 500 mapping.** `MilestoneGateError::Denied` bubbles up as a generic 500 rather than a 422/403 — the caller cannot distinguish "your gate denied this transition" from "the server is broken."
4. **`Object` schema on the gate field is unusable for SDK generation.** The TypeScript types regenerate to `Record<string, unknown>` with no structural constraint, so the SDK surface is strictly worse than having no gate field at all.

The 5-slice progression surface in #1561 gets the observable behavior (readiness, transition history, program lifecycle status) without taking on CCL-in-core as a dependency. Institution packages that want gate logic can evaluate criteria against the data exposed by `/preview` at their own boundary.

If CCL-backed gates become in-scope again later, the correct path is a separate PR that (a) versions the milestone byte layout, (b) wraps gate evaluation + status persistence in a sled transaction, (c) maps gate denial to 422, and (d) exposes `GateExpr` as a typed schema rather than `Object`.

Issue #1623, opened 2026-04-24
fix(governance): persist standalone proposals, votes, and delegations beyond domains


fahertym:
## Context

#1600 / #1621 made governance **domains** durable across gateway restart in standalone mode (the standalone path now writes through `SledGovernanceStateStore::save_domain`). This was intentionally narrow.

## Gap

`GovernanceManager` in standalone mode still keeps several institutional artifacts in plain in-memory `HashMap`s — verified at `icn/apps/governance/src/manager.rs:2697-2754`:

```rust
/// In-memory storage for proposals (standalone mode only)
proposals: RwLock<HashMap<ProposalId, Proposal>>,
/// In-memory storage for votes (standalone mode only)
votes: RwLock<HashMap<ProposalId, Vec<Vote>>>,
/// In-memory storage for delegations (standalone mode only)
delegations: RwLock<HashMap<DelegationId, Delegation>>,
```

These all reset on gateway restart. After #1621 a domain still exists, but every proposal raised against it, every vote cast on those proposals, and every delegation chain configured under it disappears.

Other in-memory state in `GovernanceManager` may be in the same boat (action items, structures, programs, milestones, meetings, agenda, discussions). Confirm during scoping.

## Why this matters

For the institutional ecosystem arc (NYCN bootstrap → live cooperative coordination), the gateway must be restartable without losing institutional history. Domains-only persistence is a partial fix.

## Suggested approach

Mirror the #1621 pattern:

1. Extend `GovernanceStateStore` (or a sibling trait) with the missing artifact write-throughs (`save_proposal`, `save_vote`, `save_delegation`, …).
2. Carry an `Option<Arc<dyn ...>>` on `GovernanceManager` set in `new_with_sled` and `None` in actor-backed constructors.
3. Seed in-memory caches from store on construction; write through on every mutator.
4. One persistence-proof integration test per artifact type.

Keep this scoped: each artifact family can be its own PR. Suggested order — proposals first (they're the spine), then votes, then delegations.

## Out of scope

- Actor-backed mode (already persisted via `SledGovernanceStateStore`).
- New artifact types — only persisting the existing in-memory ones.
- Schema migration from older standalone runs (no on-disk state existed before).

## References

- #1600, #1621 — domain persistence precedent
- `apps/governance/src/state_store.rs` — existing `SledGovernanceStateStore`
- `apps/governance/tests/standalone_domain_persistence.rs` — three-phase test pattern to mirror

PR #1624, opened 2026-04-24
feat(governance): live charter activation endpoint

### LINKED ISSUE #1602: feat(ccl): implement live charter registration/activation endpoint
NYCN bootstrap apply reports 1 "Unsupported" operation: "store charter". CCL charters validate locally but there is no generic live charter registration/activation sink.

Bootstrap apply needs a stable endpoint to hand off compiled/validated charter documents to the running runtime. This is distinct from local validation.

fahertym:
## Summary

Closes #1602.

Adds **POST /v1/gov/charters** — the bootstrap-side counterpart to the governance-ratification charter path. Lets a caller hand a validated CCL document directly to the running runtime so `CharterPolicyOracle::deploy_charter` takes effect without going through the proposal/vote/close flow.

This unblocks the NYCN bootstrap arc: `icnctl institution bootstrap apply` was reporting "store charter" as Unsupported because the only path that deployed charters was a fully-ratified `Charter` proposal. The endpoint reuses the existing `on_charter_accepted` hook (already wired by icnd to the charter oracle), so deployment goes through exactly the same code path that ratified proposals use — no parallel charter-deployment lane.

## Why

Bootstrap and governance-ratification are *both* legitimate paths for charter activation. Bootstrap charters are out-of-band ratifications (a coop's existing constitution, a federation's founding accord) — there is no proposal to vote on, the decision is already made. The endpoint accepts that semantic and gets the document into the oracle so the kernel can enforce it.

## How

- **Validation at the boundary.** Parse + structurally validate the CCL YAML in the handler. Malformed input returns 400 instead of being silently dropped by the fire-and-forget hook.
- **Reuse the existing hook.** `activate_charter` calls `ctx.on_charter_accepted` with `(charter_id, charter_yaml)` — the same hook the governance-ratification path uses. Wiring is unchanged in icnd.
- **Idempotent reactivation.** `deploy_charter` overwrites in place per `charter_id`, so re-activating the same id succeeds with a fresh `activated_at`. Callers needing single-shot semantics dedupe upstream.
- **Auth.** Requires `governance:write` scope (matches `create_domain`).
- **Regulatory-safe vocabulary.** Response uses `status`, `activated_at` — no `payment`, `currency`, or `balance` terms (asserted in tests).

### Files
- `apps/governance/Cargo.toml` — add `icn-ccl` dep for boundary validation.
- `apps/governance/src/http/models.rs` — `ActivateCharterRequest` / `ActivateCharterResponse`.
- `apps/governance/src/http/validation.rs` — `validate_charter_id` (alphanumeric/-/_/:, ≤128), `validate_charter_yaml` (non-empty, ≤1 MiB DoS guard).
- `apps/governance/src/http/handlers.rs` — `activate_charter` handler.
- `apps/governance/src/http/configure.rs` — register `POST /charters`.
- `apps/governance/tests/charter_activation.rs` — 7 integration tests.

## Risk

- **Hook ordering.** If `on_charter_accepted` is unwired the endpoint returns **500** (never silently succeeds). icnd wires it unconditionally; only test environments that opt out would see 500.
- **Idempotent overwrite.** Reactivation replaces the prior CCL document in the oracle. This matches what governance-ratification does today, but a malicious caller with `governance:write` could push a weakened charter. That risk already exists via the proposal path; the endpoint inherits the same scope check.
- **Persistence.** The oracle is in-memory. After gateway restart the charter must be re-activated. This is consistent with the governance-ratification path and not introduced by this PR.

## Test plan

- [x] `cargo fmt --all --check`
- [x] `cargo clippy -p icn-governance-actor -p icn-store -p icn-gateway --all-targets -- -D warnings`
- [x] `cargo test -p icn-governance-actor` — 7 new tests pass + existing tests unchanged.
- [x] `cargo test -p icn-store -p icn-gateway` — green.
- [x] `cargo check --workspace --all-targets` — green.

## Test cases proven

| Test | Asserts |
|---|---|
| `activate_charter_returns_201_and_fires_hook` | Valid YAML → 201 with stable identifiers, hook invoked exactly once with submitted (id, yaml), regulatory-safe vocabulary in response |
| `activate_charter_idempotent_on_reactivation` | Same charter_id twice → both 201 |
| `activate_charter_rejects_malformed_yaml` | Bad YAML → 400, hook NOT invoked |
| `activate_charter_rejects_empty_id` | empty charter_id → 400 |
| `activate_charter_rejects_empty_yaml` | whitespace-only YAML → 400 |
| `activate_charter_requires_governance_write_scope` | Missing claims → 401/403, hook NOT invoked |
| `activate_charter_returns_500_when_hook_not_wired` | Unwired hook → 500 (never silent success) |

## Follow-ups

- #1623 (separate issue) — Standalone-mode persistence for proposals/votes/delegations. Not in scope here.
- OpenAPI/TypeScript types regenerate on the next SDK refresh; no manual regen needed in this PR.

### DIFF
--- a/icn/Cargo.lock
+++ b/icn/Cargo.lock
@@ -3284,6 +3284,7 @@ dependencies = [
  "blake3",
  "ed25519-dalek",
  "hex",
+ "icn-ccl",
  "icn-encoding",
  "icn-entity",
  "icn-federation",
--- a/icn/apps/governance/Cargo.toml
+++ b/icn/apps/governance/Cargo.toml
@@ -13,6 +13,9 @@ icn-kernel-api = { path = "../../crates/icn-kernel-api" }
 # Governance domain types (proposals, votes, domains, etc.)
 icn-governance = { path = "../../crates/icn-governance" }
 
+# CCL schema (charter YAML parsing/validation at the HTTP boundary)
+icn-ccl = { path = "../../crates/icn-ccl" }
+
 # Identity types (Did)
 icn-identity = { path = "../../crates/icn-identity" }
 
--- a/icn/apps/governance/src/http/configure.rs
+++ b/icn/apps/governance/src/http/configure.rs
@@ -295,6 +295,8 @@ where
                 .route(web::post().to(handlers::add_domain_member::<E>))
                 .route(web::delete().to(handlers::remove_domain_member::<E>)),
         )
+        // ── Charter activation ────────────────────────────────────────────
+        .service(web::resource("/charters").route(web::post().to(handlers::activate_charter::<E>)))
         // ── Proposal endpoints ────────────────────────────────────────────
         .service(
             web::resource("/proposals")
--- a/icn/apps/governance/src/http/handlers.rs
+++ b/icn/apps/governance/src/http/handlers.rs
@@ -638,18 +638,59 @@ pub async fn remove_domain_member<E: GovernanceEventEmitter + Clone + 'static>(
 // Charter activation handlers
 // ============================================================================
 
+/// Stable, recognizable proposal_id used when the direct charter-activation
+/// path emits a [`GovernanceEffect::DeployCharter`]. The dispatch site
+/// (`server.rs`) keys commons registration on `charter_id` and ignores
+/// `proposal_id`, so a synthetic value is safe and makes the provenance
+/// origin obvious in logs.
+fn direct_activation_pseudo_proposal_id(charter_id: &str) -> String {
+    format!("direct-activation:{charter_id}")
+}
+
 /// POST /gov/charters — Activate a CCL charter directly.
 ///
 /// Bootstrap-side counterpart to the governance-ratification path: hands a
-/// validated CCL document straight to the gateway's `on_charter_accepted`
-/// hook (which deploys it into the `CharterPolicyOracle`).  Does **not**
-/// create a proposal, record a vote, or emit a decision — the charter is
-/// treated as already-decided by an out-of-band ratification.
+/// validated CCL document to the same downstream pair the accepted-proposal
+/// path uses.
 ///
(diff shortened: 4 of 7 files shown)


fahertym:
## Review feedback addressed (commit c7470e95)

1. **P1 — Direct activation now dispatches `GovernanceEffect::DeployCharter`.** The handler invokes both hooks, mirroring `close_proposal`'s charter branch. Without this, commons would not register the domain and SDIS `register_steward` would reject the jurisdiction as unknown. Synthetic `proposal_id = direct-activation:<charter_id>` (the gateway dispatch keys on `charter_id` and ignores `proposal_id`).

2. If `on_proposal_accepted` is unwired the handler now returns **500** instead of silently succeeding. A 201 response now means the charter is fully activated (oracle + commons), not half-activated.

3. `MAX_CHARTER_YAML_BYTES` lowered from 1 MiB to **256 KiB** to match the gateway-wide `web::JsonConfig::default().limit(262_144)` in `icn-gateway/src/server.rs`.

4. Module doc comment now matches what is actually asserted. Added `activate_charter_rejects_oversize_yaml` and `activate_charter_returns_500_when_proposal_hook_not_wired`. Existing tests also now assert the DeployCharter dispatch.

**Out of scope (documented in code):** institutional effect record / decision-hash provenance for direct activation. The ratified path emits an `InstitutionalEffectRecord` keyed by a real `proposal_id` bound to a governance receipt. Direct activation has no proposal, so provenance for this path is the tracing audit trail plus the synthetic `proposal_id` in commons logs. A synthetic proposal record would require a broader institutional-history redesign — tracked elsewhere.

**Tests:** 9/9 charter_activation pass; 219 governance-actor unit tests pass; gateway tests pass; fmt/clippy clean.

Issue #1603, opened 2026-04-24
feat(bootstrap): support role assignment in bootstrap seeds with real person DIDs


fahertym:
4 role assignments were deferred during live NYCN bootstrap apply: coordinator on nycn-steering, treasurer on nycn-finance, program-coordinator on nycn-content, operations-lead on nycn-logistics. These were deferred because no `person_did` was supplied in the seed data.

Need a mechanism to supply real person DIDs in bootstrap package seeds and wire them through apply.

PR #1627, opened 2026-04-25
feat(governance): add GET /me/standing read model

### LINKED ISSUE #1604: feat(gateway): implement GET /me/standing — identity, memberships, current scopes
First implementation slice from `docs/architecture/MEMBER_STANDING.md`. Returns: caller DID, individual entity ID if present, display label if available, memberships (entity ID/type/role/status), and current scopes from `/gov/me/scopes`.

This is the minimum viable standing response needed as substrate for Commons Shell, Action Cards, and accessible governance participation.

fahertym:
## Summary

Closes #1604.

Adds the caller-facing institutional read model — `GET /v1/gov/me/standing` — that answers *"what is my standing in this institution/domain/federation?"* by composing existing governance state (domain memberships + structure role assignments) into one digestible response.

This is the first piece of the member-facing read surface that NYCN-style federations and other institution packages will dogfood. It is intentionally generic: the endpoint knows nothing about NYCN.

## What `/me/standing` returns

`StandingResponse`:

| Field | Meaning |
|---|---|
| `did` | Caller's authenticated DID. Always derived from claims, never from a query param. |
| `display_label?` | Reserved for a future identity/entity-registry hop. Always `None` here. |
| `domains[]` | Governance domains where the caller is in the static member list, plus trust-threshold domains marked `"unverified"` (the trust graph is the source of truth, not this read model). |
| `roles[]` | Role assignments held by the caller, joined with cheap structure metadata (name + parent entity) so UI does not need a per-row fetch. |
| `authority_scopes[]` | Deduplicated, sorted union of `authority_scope` across roles. Convenience for UI. |
| `generated_at` | Unix epoch seconds when the snapshot was taken. No server-side caching. |

Optional filter: `?domain_id=<id>` narrows both `domains` and `roles` to that domain. Unknown ids return **200 with empty arrays** (not 404) — *"no standing in that domain"* is a valid answer.

## How it uses existing DID/auth context

- Mounted under the existing `/v1/gov` scope, alongside `/me/scopes` and `/me/work`.
- Requires `governance:read` (read-only) — matches `/me/scopes`. Reading your own standing must never require write authority.
- Caller DID = `claims.sub`. There is no `did=` query param, so a caller cannot inspect another DID's standing. Pinned by test `caller_does_not_see_another_dids_standing`.

## How this connects to #1603 person-directory role assignment

The person-directory overlay merged in #1626 resolves `holder_label → person_did` at plan time, then apply calls the same `manager.assign_role` API the inline-DID path uses. `/me/standing` reads from the same `StructureStore` both paths populate, so directory-resolved roles and inline-DID roles surface identically — no separate code path. Pinned by tests `person_directory_resolved_role_lands_the_same_way` and `inline_did_precedence_from_1603_is_preserved_in_standing`.

## How this prepares for NYCN deployment without hardcoding NYCN

NYCN ships a public package (structures, programs, role-assignment seeds with `holder_label`) and a private overlay (`person-directory-seed` mapping labels to real DIDs). After `apply` runs:

- A NYCN organizer hits `/me/standing` and sees their committee membership, their assigned roles, and the structures they belong to — without any NYCN-specific code in ICN core.
- An unrelated co-op operator using a different package hits the same endpoint and sees their own institution's standing the same way.

The endpoint is generic over institution; what changes between deployments is the seed data, not the runtime.

## Why action cards are intentionally deferred

`/me/standing` is the substrate. Action-card generation ("here are 3 things you can do right now") joins this read model with proposal/meeting/work state and applies UX-layer prioritization. That belongs in a separate endpoint and PR — bundling them now would couple a stable read model to a still-evolving UX pattern.

## Files

- `icn/apps/governance/src/http/models.rs` — `StandingQuery`, `StandingResponse`, `StandingDomainMembership`, `StandingRoleAssignment` (typed structs, not ad-hoc JSON).
- `icn/apps/governance/src/http/handlers.rs` — `get_my_standing` handler.
- `icn/apps/governance/src/http/configure.rs` — register `GET /me/standing`.
- `icn/apps/governance/tests/me_standing.rs` — 9 integration tests.

## Test plan

- [x] `cargo test -p icn-governance-actor --test me_standing` — 9/9 pass.
- [x] `cargo test -p icn-governance-actor` — full suite green (no regressions).
- [x] `cargo check -p icn-gateway --all-targets` — gateway re-binds the route group cleanly.
- [x] `cargo fmt --all --check` — clean.
- [x] `cargo clippy -p icn-governance-actor --all-targets -- -D warnings` — clean.

## Test cases proven

| Test | Asserts |
|---|---|
| `caller_with_no_standing_returns_well_formed_empty` | Empty caller → 200 with empty arrays, never errors |
| `direct_role_assignment_surfaces_in_standing` | Manager-level role assignment shows up with structure metadata joined |
| `person_directory_resolved_role_lands_the_same_way` | #1603 overlay path produces identical runtime visibility |
| `inline_did_precedence_from_1603_is_preserved_in_standing` | Inline DID wins; directory DID does NOT inherit standing |
| `caller_does_not_see_another_dids_standing` | Self-only — no `did=` param, no cross-DID visibility |
| `domain_membership_static_list_surfaces` | Static-list domain membership rolls into `domains[]` |
| `domain_filter_narrows_results_and_unknown_id_returns_empty` | `?domain_id=` filters and unknown id is 200-empty, not 404 |
| `response_uses_regulatory_safe_vocabulary` | No `payment` / `currency` / `balance` terms |
| `missing_governance_read_scope_is_rejected` | 401 or 403 on missing scope |

## Limitations / known gaps (documented in code)

- **`display_label`** is always `None` until an identity-registry lookup is wired. That's a separate slice when the entity-registry read API stabilizes.
- **`authority_scopes`** rolls up empty in the end-to-end path because `manager.assign_role` and `POST /v1/gov/structures/{id}/roles` do not currently accept an `authority_scope` parameter — the field exists on `BootstrapRoleAssignmentRecord` and on `RoleAssignment` but is dropped at apply time. Plumbing it through is a small follow-up tracked alongside #1603. The rollup logic itself (sort + dedup) is structurally correct; the input vector is just empty today.
- **Domain filter** currently joins `roles` to a domain by matching the structure's `parent_entity_id` against the requested `domain_id`. This is the loose-but-correct join until structures carry an explicit governance-domain back-reference. Sufficient for current institution packages where entity_id ≈ governance_domain_id.

## Recommended next follow-up

1. **NYCN deploy/apply importer path** — package operator workflow that takes a public package + private overlay and runs `icnctl institution bootstrap apply` end-to-end against a deployed gateway. With `/me/standing` available, organizers can verify their assignments landed.
2. **Action cards on top of `/me/standing`** — once standing returns real data for real organizers, layer the action-card generator that surfaces "here are 3 things you can do right now."
3. **`authority_scope` plumbing** — extend `assign_role` and the HTTP request type to carry the field bootstrap planning already produces.

### DIFF
--- a/docs/architecture/INSTITUTION_PACKAGE_BOUNDARY.md
+++ b/docs/architecture/INSTITUTION_PACKAGE_BOUNDARY.md
@@ -115,6 +115,103 @@ CCL describes *what is permitted and under what conditions*. The host runtime de
 
 ---
 
+## C2. Entity Primitives vs Local Purpose Metadata (the type/purpose firewall)
+
+**The same rule that separates kernel from app applies to entity types.**
+
+ICN has exactly **four entity primitives**. They are sovereign types and they imply different machinery:
+
+| Primitive | What it is | What ICN guarantees |
+|-----------|------------|---------------------|
+| `Individual` | A DID-backed person. | Identity, authentication, self-scoped reads. |
+| `Cooperative` | An operating/economic body. | `governance_domain_id` + `treasury_account` (per `CooperativeProfile`). |
+| `Community` | A civic/member/participant body. | `governance_domain_id` (per `CommunityProfile`). Treasury optional. |
+| `Federation` | An association of member entities. | `governance_domain_id` + `member_entities` + optional `parent_id` (per `FederationProfile`). |
+
+These are the only legal values of `BootstrapEntityType` in `icn-governance::bootstrap` and the only legal values of the `entity_type` field on the gateway's `/v1/entities` API. The serde enum is `#[serde(rename_all = "snake_case")]` and any other variant fails to deserialize at YAML parse time. **This is the firewall**: an institution package cannot smuggle a new entity type past ICN.
+
+**Committees and working groups are NOT entities.** They are governance `Structure` records owned by a parent entity. They carry delegated authority via `RoleAssignment.authority_scope`, not sovereignty. They cannot hold treasury, cannot join federations, cannot appear in `member_entities`.
+
+**Events / programs / cycles (e.g. a Summit) are NOT entities.** They are `Activity` and `Program` records owned by a parent entity.
+
+### Local institutional role goes in `operating_purpose`, not in a new type
+
+Institution packages frequently want to mark *which kind* of cooperative or community a record represents in their local taxonomy: an "organizing cooperative", a "program participant community", a "federation member community". These are valid local distinctions — but they are **purpose metadata**, not entity primitives.
+
+`BootstrapEntityRecord` already carries:
+
+```rust
+pub operating_purpose: Option<String>,
+```
+
+That field is the only sanctioned home for this metadata. It is opaque to ICN — the runtime stores it (or, in the current importer, plans for it; see "Known gap" below) and never branches on its value.
+
+### Correct vs incorrect mapping
+
+```yaml
+# ❌ WRONG — invents new entity primitives, will fail to parse
+entity_type: organizing_cooperative
+entity_type: program_community
+entity_type: committee
+entity_type: federation_member_community
+
+# ✅ RIGHT — uses ICN primitives, locales role via operating_purpose
+entity_type: cooperative
+operating_purpose: organizing_cooperative
+
+entity_type: community
+operating_purpose: program_participant_community
+# (with a separate Activity record for the related program)
+
+# Committee — not an entity at all:
+# Goes in a `kind: structure-seed` manifest:
+kind: structure-seed
+parent_entity_id: nycn-organizing
+structures:
+  - id: committee-finance
+    kind: committee
+    name: Finance Committee
+```
+
+### Worked example: NYCN reference federation
+
+```text
+nycn                        entity_type: federation
+                            operating_purpose: regional_cooperative_network
+
+  nycn-organizing           entity_type: cooperative
+                            operating_purpose: organizing_cooperative
+                            (carries treasury_account)
+
+    committee-steering      structure-seed { kind: committee,  parent_entity_id: nycn-organizing }
+    committee-finance       structure-seed { kind: committee,  parent_entity_id: nycn-organizing }
+    committee-program       structure-seed { kind: committee,  parent_entity_id: nycn-organizing }
+    wg-accessibility        structure-seed { kind: working_group, parent_entity_id: nycn-organizing }
+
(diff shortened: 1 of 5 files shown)


fahertym:
## Review feedback addressed (commit ac68c10f)

1. **handlers.rs** — `get_structure` errors now propagate via `map_err(anyhow_to_api)` instead of being swallowed by `.ok().flatten()`. A genuine `Ok(None)` (deleted structure) still surfaces a row with `structure_name: None` so the caller can investigate; an `Err` becomes 500. Silently masking storage failures as "structure deleted" was the concrete risk Copilot flagged.

2. **me_standing.rs** — `response_uses_regulatory_safe_vocabulary` now asserts `StatusCode::OK` before scanning the body. An auth or routing regression returning 401/403/500 would otherwise let a clean error payload pass the term check.

## Documentation pinned

Added section **C2 "Entity Primitives vs Local Purpose Metadata"** to `docs/architecture/INSTITUTION_PACKAGE_BOUNDARY.md`. This pins the type/purpose firewall:

- The four ICN entity primitives are `Individual`, `Cooperative`, `Community`, `Federation`. Anything else fails to deserialize at YAML parse time.
- Committees / working groups are governance `Structure` records under a parent entity, not entities.
- Events / programs / cycles (Summit) are `Activity` / `Program` records, not entities.
- Local institutional role goes in `BootstrapEntityRecord.operating_purpose` — opaque to ICN.
- Worked NYCN example included.

This addresses the ontology error where institution packages were authoring `entity_type: organizing_cooperative`, `entity_type: program_community`, or `entity_type: committee`. The bootstrap importer schema already rejects those mechanically (serde refuses to parse), but the doc now states *why* and points authors at `operating_purpose`.

## Follow-up issues opened

- **#1628** — `feat(bootstrap): forward operating_purpose to /v1/entities at apply time`. The seed schema accepts `operating_purpose`, the validator/planner sees it, but `BootstrapOperation::CreateEntity` drops it before the apply step POSTs to `/v1/entities`.
- **#1629** — `feat(governance): plumb authority_scope through assign_role end-to-end`. `AssignRoleRequest` and `manager.assign_role` don't carry `authority_scope`, so `/me/standing`'s `authority_scopes` rolls up empty even after #1603's directory resolution lands a real DID.

Both are narrow plumbing slices — intentionally out of scope for #1627.

PR #1661, opened 2026-04-27
feat(governance): add action item completion receipts

fahertym:


fahertym:
Thanks — all five threads addressed in [\`cfbda92b\`](https://github.com/InterCooperative-Network/icn/commit/cfbda92b).

**1. Append-only completion-receipt history (Codex P1)** — `put_action_item_completion` now writes a primary record keyed by `record_hash` plus a `by_item` secondary index keyed by `(u64 item_id_len, item_id, u64 completed_at, record_hash)`. A reopen / re-complete cycle produces a distinct `record_hash` (because `completed_at` advances) and a distinct primary record; the prior receipt is preserved. New trait method \`list_action_item_completions_by_item\` returns the full chain oldest-first; default \`Ok(vec![])\`. \`get_action_item_completion_by_item\` returns the latest by `completed_at`. New test: \`reopen_and_recomplete_preserves_completion_history\`.

**2. Fail completion when receipt persistence fails (Codex P1)** — \`update_action_item_status\` now persists the receipt BEFORE saving the status transition. Backend rejection propagates as an Err and the action-item save does not run. New test: \`receipt_backend_failure_prevents_completed_status_commit\` wires a \`FailingCompletionStore\` and asserts the item's stored status stays at the prior value.

**3. Full-update handler bypass (Copilot)** — the full-update handler (\`PUT /domains/{domain_id}/action-items/{item_id}\`) now applies non-status fields via \`manager.update_action_item\`, then routes any status change through \`manager.update_action_item_status\` with the authenticated DID as actor. The \`/status\` endpoint is no longer the only completion path; both paths now produce the same receipt. New test: \`full_update_to_completed_emits_completion_receipt\` exercises the HTTP handler and asserts the receipt persists.

**4. Canonical hash length prefixes (Copilot)** — switched the three string field length prefixes in \`ActionItemCompletionReceipt::compute_record_hash\` from \`u32\` to \`u64\`. Now consistent with \`GovernanceProof::compute_proof_hash\` and \`compute_vote_hash\` elsewhere in the module.

**5. Per-write sled flush (Copilot)** — removed the per-write \`db.flush()\` from \`ReceiptStore::put_action_item_completion\`. Durability semantics now match \`put_governance\` and the other receipt writes.

Local re-validation:
- \`cargo fmt --all --check\` ✅
- \`cargo clippy -p icn-governance-actor -p icn-governance -p icn-gateway --all-targets -- -D warnings\` ✅
- \`me_action_item_receipt_chain\` 6/6 (was 3) — three new tests for full-update routing, backend-failure abort, append-only chain.
- \`me_action_card_receipt_chain\` 3/3, \`me_action_cards\` 4/4, \`me_standing\` 9/9 — no regression.
- \`icn-governance-actor --lib\` 220/220, \`icn-governance --lib\` 502/502, \`icn-gateway --lib\` 528/528.

PR #1775, opened 2026-05-08
demo(pilot-ui): align ?mode=demo shell with organizing-committee path

fahertym:
## Why

Manual inspection of `http://localhost:8765/?mode=demo` after #1774 merged showed the visual shell still framed the demo as **ICN Timebank**:

- Browser tab and in-app header read **ICN Timebank**.
- First-class nav showed **Dashboard / Log Hours / History / Members / Profile** above-the-fold.
- The Demo Guide still listed standing, action cards, and plain-language receipts as **Coming next in this demo tranche — Not yet wired (PR 3 / PR 4 / PR 5)**.

After #1771–#1774, those PR-tranche labels are stale. Standing, action cards, and the receipt-summary shell shipped. The demo is the ICN organizing-committee / civic coordination demo, not a timebank demo with extras bolted on. The visual shell needs to match.

## What changed (demo mode only — non-demo unchanged)

- **Retitle** the visible app to `ICN Organizing Committee Demo` — `document.title` (browser tab), in-app header `<h1>`, and login-screen `<h1>`. `applyDemoMode()` does the swap at script-eval time. The static HTML defaults stay `ICN Timebank` so the non-demo / live timebank surface keeps its existing identity until a global rename is explicitly requested.

- **De-emphasize timebank nav** in demo mode: `Dashboard`, `Log Hours`, `History`, `Members`, `Profile` carry `data-demo-irrelevant="true"` and are hidden via the existing `.hidden` class when `?mode=demo`. Preferred demo-mode nav becomes:
  - `Demo Guide`
  - `My Standing & Action Cards`
  - `Governance`
  - `Receipts`
  - `Federation` (kept; honestly labeled in the Demo Guide as gateway-backed, not fixture-backed)
  
  The tab DOM is untouched, so anchor / programmatic `switchTab()` into any tab still works for adjacent tools.

- **Banner copy**: drop the stale "Gateway-side fixture/demo mode arrives in PR 5." line. New copy reads:
  > Local organizer/member demo. Standing and action cards are fixture-backed; governance, receipts, ledger, members, trust, and federation are not fixture-backed yet.

- **Demo Guide rewrite**: replace the old "Coming next" / PR3/PR4/PR5 block with truth-current sections matching the §3 Guided workflow narrative:
  - Available now in this local demo
  - **Fixture-backed today** (standing, action cards)
  - **Visible but still gateway-backed / not fixture-backed** (governance, receipt-chain, ledger, members, trust, federation)
[...]

PR #1776, opened 2026-05-08
demo(pilot-ui): improve local demo guide visual hierarchy

fahertym:
## Why

Manual visual inspection of `http://localhost:8765/?mode=demo` after #1775 merged showed the Demo Guide rendered as a wall of unstyled text. The page-source/server side was correct (cards, sections, copy) but nothing was actually styled.

**Root cause:** a pre-existing CSS parse error in `style.css` at line 5670 — the `#wizard-did-display` rule was missing its closing `}` and the comment that followed was missing its `/*` opener. The browser CSS parser silently dropped every rule after that point, including the demo-guide selectors landed in #1770 / #1771 / #1775.

```css
#wizard-did-display {
    ...
    border-radius: 4px;
   Inline Receipt Chain (Proposal Detail Sidebar) — P0-T11   /* never closed */
   ============================================================================ */
```

Verified via the browser CSS Object Model: stylesheet parsed **704 rules** before the fix, **835 rules** after.

### Pre-existing parse-error fix
Closed `#wizard-did-display { … }` properly and reopened the next section comment with `/*`. This is the root cause; without it, no demo-guide CSS would apply.

### Demo Guide visual hierarchy
- Hero block with eyebrow (`Local demo · Civic coordination substrate`), gradient background, primary-color left accent.
- Each existing section becomes a `.demo-guide-card` with a `data-section` attribute that drives a per-section accent color and badge:

| Section | Accent | Badge |
|--|--|--|
| Where you are | primary blue | `Context` |
| Available now in this local demo | green | `Live` |
| Fixture-backed today | purple | `Fixture` |
| Visible but still gateway-backed | amber | `Gateway-backed` |
| Not implemented in this demo | gray (dimmed) | `Out of scope` |
| Future narrow slices | cyan | `Planned` |
| Non-claims | neutral (dimmed) | `Read carefully` |

- Responsive `auto-fit, minmax(22rem, 1fr)` grid: two columns on wide viewports, one on narrow.
- Status pills on individual list items: `.demo-status--live` (green) and `.demo-status--fixture` (purple).
- "Not implemented" items get a quiet em-dash marker instead of an alarming `✕`; muted text so the section reads as out-of-scope rather than failure.
- Narrow-screen tightening at `≤720px`: single column, tighter hero, where-you-are collapses to one column.
- Light + dark theme support via existing CSS variables.

### Tag fixes
- `.demo-guide-hero` and `.demo-guide-card-header` were originally `<header>`. The global `header { display: flex; justify-content: space-between; … }` rule (line 625) hijacked them and split children across the row. Switched to `<div>`.
- Added `?v=2026050802` to the `<link rel=\"stylesheet\">` so the PWA service worker stops serving the old CSS to existing users.

## What did not change

- **No new fixtures, no backend, no runtime mode, no issue edits, no Rust touched.**
- Demo-mode shell behavior from #1775 (title swap, nav hiding, banner copy, no-gateway bootstrap) is unchanged.
- Demo Guide copy is unchanged in meaning (only one tiny wording tweak: removed the now-redundant \"Logged-in DID\" note since the hero already implies it).
- Non-demo path is unchanged.

## Validation

- `node --check` on `app.js` and both e2e specs — clean
- `python3 -m json.tool` on both fixture JSON files — clean (unchanged)
- No Rust / Cargo files in diff
- Leak / overclaim grep — only legitimate non-claim contexts hit
- `git diff --check` — clean
- Fresh-load Playwright smoke test on `http://localhost:8765/?mode=demo` — 6 styled cards rendered, distinct accents, working status pills, no broken hero
- CSS parser sanity — **835 rules parsed (up from 704)** confirms the parse-error fix landed

Existing test selectors (`.demo-guide-non-claims`, `.demo-guide-gateway-backed`, `.demo-guide-fixture`, `.demo-guide-future`, `#demo-guide-mode`, `#demo-guide-heading`, `aria-labelledby=\"demo-guide-nav-btn\"`) remain valid; classes/ids are preserved on the existing list/heading elements that are now wrapped in cards.

## Local inspection

```bash
cd /home/matt/projects/icn/web/pilot-ui
python3 -m http.server 8765
# http://localhost:8765/?mode=demo
```

If you have an old session running, hard-reload (Ctrl+Shift+R) — the `?v=2026050802` query on the new stylesheet link bypasses the service-worker cache.

## Recommended next

Pause here and let Matt visually inspect again before starting the governance proposal/vote fixture slice.

The tool-library demo remains dev scaffolding, not the ICN demo.

### DIFF
--- a/web/pilot-ui/app.js
+++ b/web/pilot-ui/app.js
@@ -29,7 +29,7 @@ const DEMO_MODE = (() => {
     }
 })();
 const DEMO_MODE_MESSAGES = {
-    demo:    'Local organizer/member demo. Standing and action cards are fixture-backed; governance, receipts, ledger, members, trust, and federation are not fixture-backed yet.',
+    demo:    'Local organizer/member demo. Only Standing and Action Cards are fixture-backed today.',
     fixture: 'UI is in FIXTURE mode — committed test data only; not real participant state.',
     devnet:  'UI is in DEVNET mode — local 3-node Docker cluster; not production.',
     local:   'UI is in LOCAL mode — single-node local gateway.',
@@ -70,6 +70,10 @@ function applyDemoMode() {
     // direct hash-link still works for adjacent tools.
     if (DEMO_MODE === 'demo') {
         try { document.title = DEMO_TITLE; } catch (_) { /* noop */ }
+        // Mark body so demo-mode-only CSS (wider main, action-cards
+        // grid, etc.) can target the right scope without leaking
+        // into non-demo pages.
+        if (document.body) document.body.setAttribute('data-demo-mode', 'true');
         const headerTitle = document.getElementById('app-title-header');
         if (headerTitle) headerTitle.textContent = DEMO_TITLE;
         const loginTitle = document.getElementById('app-title-login');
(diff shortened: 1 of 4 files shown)


fahertym:
Addressed inline review: bumped `CACHE_VERSION` 1.0 → 1.1 in `sw.js`. The existing `activate` handler already purges caches whose name doesn't match the current version, so existing installs will fetch the new bundle (style.css with the parser fix, index.html with the new card markup) on next load. Smallest fix; no fetch-strategy or static-asset-list changes needed.

fahertym:
Follow-up for the unresolved service-worker cache review thread:

Addressed in `148a42d`: bumped `CACHE_VERSION` to `icn-timebank-v1.2` and added `/style.css?v=2026050802` to `STATIC_ASSETS` so the versioned stylesheet is precached by the updated service worker while old `icn-timebank-*` caches are purged on activation.

This keeps the cache-bust and offline/static cache behavior aligned with the CSS fix without widening the PR beyond pilot-ui demo/static-asset handling.

PR #1906, opened 2026-05-24
fix(compute): use public exports in cost doctest

fahertym:
## Summary

Fixes the existing `icn-compute` cost doctest baseline failure by using the public crate-root exports for `ComputeTask` and `FuelLimit`, and by making the doctest construct a valid task without relying on a nonexistent `Default` implementation.

This does not touch runtime code.

## Testing

- `cd icn && cargo test -p icn-compute --doc`
- `cargo fmt --all --check`
- `cargo clippy --all-targets`
- `cargo test`

## Invariants

- Runtime behavior unchanged
- Protocol paths unchanged
- Canonical encodings unchanged
- Trust gates unchanged
- Kernel/app boundaries unchanged

### DIFF
--- a/icn/crates/icn-compute/src/cost.rs
+++ b/icn/crates/icn-compute/src/cost.rs
@@ -34,28 +34,49 @@ const DEFAULT_FUEL_COST_DIVISOR: i64 = 1_000;
 /// # Examples
 ///
 /// ```rust,no_run
-/// use icn_compute::types::{ComputeTask, FuelLimit};
 /// use icn_compute::cost::compute_credits_required;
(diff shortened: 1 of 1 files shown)


fahertym:
Acknowledging the benchmark comparison with `perf-regression-ok`: this PR changes only the `icn-compute` cost doctest/example block and does not modify runtime code. Required protection checks are green (`Build Release`, `Test`, `Clippy`, `Format Check`, firewall checks, TypeScript SDK, Accessibility Tests, Regulatory Compliance Linter, and Agent Tooling Drift Check`).

PR #1952, opened 2026-05-30
docs(summit): align workshop proposal claims to verified runtime; fold in proven loop

fahertym:
## What
Revises `docs/summit/workshop-proposal-draft.md` (Oct 2026 NY Coop Summit) to submit-ready, and **aligns its claims to what ICN verifiably does today**.

## Why
The prior draft promised a live governance **vote** and five flows on the K3s cluster. Two problems: (1) the `propose -> vote -> decision-receipt` path is gated on the commons **Member-standing** layer (enroll -> voucher -> join jurisdiction -> advance to Member), which is in development — it can't be staged live without that subsystem; (2) the cluster currently has crashlooping coop pods. Shipping those claims would overstate readiness — the opposite of what the proposal is trying to earn from organizers.

## How
- **Lead the live demo with the proven loop** (§3a): cooperative work → legible obligation → **verifiable completion receipt** (`record_hash`). Verified end-to-end on current `main`, 2026-05-30, on a local node (reproducible script + transcript exist).
- **Reframe voting as in-development** (architecture + fixtures, not staged live).
- **Drop the hard cluster dependency**: proven loop runs on a laptop; cluster is "the same software at scale" for the fuller-picture scenarios (§3b), each labeled by status.
- **Add a "proven / rehearsed / next" truth box** shown to attendees — the honesty *is* the pitch.
- **Add a slide outline (Appendix A) and one-page handout outline (Appendix B).**

## Not in scope
Presenter voice pass and the Summit call's exact deadline/submission channel are the author's to finalize. No runtime/code change; docs only.

### DIFF
--- a/docs/summit/workshop-proposal-draft.md
+++ b/docs/summit/workshop-proposal-draft.md
@@ -1,7 +1,7 @@
 # Workshop Proposal: Democratic Infrastructure for Cooperatives
 
 **Submitted to:** New York Cooperative Summit — October 2026 Call for Presenters
-**Draft date:** 2026-03-23
+**Draft date:** 2026-03-23 · **Revised:** 2026-05-30 (claims aligned to verified runtime; proven live loop folded in)
 **Track:** Technology & Infrastructure
 **Format:** Workshop (90 minutes)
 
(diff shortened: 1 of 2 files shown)


fahertym:
Addressed in 2217a531: (1) narrowed the completion-receipt claim to the fields `compute_record_hash` actually binds (item_id, domain_id, actor_did, transition, completed_at) — it proves the completion *event*, not the editable title/description text; (2) synced the package-local copy `institutions/nycn/summit/docs/workshop-proposal-draft.md` (was identical to the old draft) to the aligned content, so there are no longer two conflicting submit-ready proposals. Single-sourcing the two copies is noted as a follow-up.

PR #2492, opened 2026-08-01
chore(build): exclude the Rust target dir from the Docker build context

fahertym:
## What

Adds `icn/target/` and `**/target/` to `.dockerignore`.

## Why

`.dockerignore` already listed `target/`, but in `.dockerignore` a bare `target/` matches only the **build-context root**. The Cargo workspace lives in `icn/`, and the file then re-includes the whole tree with `!icn/` — so `icn/target/` was never excluded.

Every image build has been tarring it into the context. On the dev VM that directory is **38G**: builds sat for 20+ minutes without reaching Step 1, and emitted a stream of

```
Can't add file .../icn/target/debug/deps/<hash>.rcgu.o to tar:
lstat ...: no such file or directory
```

as a concurrently-running `cargo` deleted files mid-tar — which also makes the context nondeterministic depending on what else is running.

Placement matters: the new rules go **after** `!icn/`, because the last matching rule wins. Earlier in the file they would simply be undone by the re-include.

## Risk

None to the built artifact. The Dockerfile compiles from source inside the builder stage (`RUN cargo build --release …`); nothing in `target/` is consumed. Excluding it only removes host build artifacts that could never be used, and removes a source of nondeterminism.

## Test plan

Observed directly, same command both times:

- Before: `docker build` stalled in context upload past 20 minutes, never reaching Step 1, with repeated `Can't add file … to tar` errors.
- After: reaches `Step 15/26 : RUN cargo build --release …` immediately and runs to `Successfully tagged icn:828eb596`, exit 0.

The resulting image was deployed to all four coop nodes and verified running.

Found while deploying #2490; kept separate because it is build infrastructure rather than a code fix.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -20,3 +20,10 @@ monitoring/
 !config/
 !contracts/
 !scripts/
+
+# Rust build artifacts. The bare `target/` rule above only matches the context
+# root, and `!icn/` re-includes the whole workspace tree — so without this the
+# 38G icn/target/ directory is tarred into the build context on every build.
+# Must come after `!icn/`: in .dockerignore the last matching rule wins.
+icn/target/
+**/target/
--- a/.dockerignore
+++ b/.dockerignore
@@ -27,3 +27,12 @@ monitoring/
 # Must come after `!icn/`: in .dockerignore the last matching rule wins.
 icn/target/
 **/target/
+
+# Exception: deploy/Dockerfile.icnd-fast copies these two host-built binaries
+# straight from the build context (see demo/RUNBOOK.md), so they must survive
+# the exclusion above. Only these two paths are re-included — the rest of
+# target/ (tens of GB of intermediate artifacts) stays out.
+# deploy/Dockerfile.icnd is unaffected either way: it uses COPY --from=builder,
+# which reads from the build stage rather than the context.
+!icn/target/release/icnd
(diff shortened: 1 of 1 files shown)


fahertym:
Correct catch — fixed in `670e12ab`.

`deploy/Dockerfile.icnd-fast:8-9` does copy `icn/target/release/{icnd,icnctl}` straight from the build context, and `demo/RUNBOOK.md:229` documents that path, so the previous version of this change would have failed both `COPY` instructions. I verified the Dockerfile rather than taking the report at face value, and it is exactly as described.

`deploy/Dockerfile.icnd` is unaffected either way — it uses `COPY --from=builder`, which reads from the build stage, not the context.

The fix re-includes only those two paths. I checked the behaviour in an isolated fixture rather than assuming Docker re-includes files from an excluded directory:

- `icn/target/release/{icnd,icnctl}` — present in the context, copy cleanly
- `icn/crates/` source — still present
- **negative control:** `icn/target/release/libfoo.rlib`, sitting in the *same directory* but not excepted, fails with `COPY failed: file not found in build context or excluded by .dockerignore`

So the exception is narrow rather than a re-include of the parent directory. Context size in the fixture was 8.192kB with a 2MB file planted in `target/debug/`.

On the real repo the build context is now **30.61MB**, against a 38G target tree.

Issue #2517, opened 2026-08-03
fix(net): the durable signing sequence resumes below the pre-upgrade high-water peers already recorded


fahertym:
## Symptom (live, right now)

Gamma is rejecting **legitimate** traffic from beta on the rehearsal federation:

```
Replay attack detected from did:icn:zDMiXkUafnaRf... (beta, 10.42.2.206)
  sequence 12901 already seen (max: 15915)
  sequence 12974 already seen (max: 15915)
```

39 rejections in 60 minutes and ongoing.

## This is not #2514

Note the message says **`max:`**, not `floor:` — this is the Bloom-filter branch of `check_replay_only`, not the restart-floor branch. #2514's `+1000` gap is a different mechanism with a different symptom.

## Cause

`SigningSequenceCounter` (#2510 / #2511) made the outbound sequence durable. On beta's upgrade it initialised fresh:

```
icn_net::signing_sequence: Loaded durable signing sequence counter resumed_at=10001 block=1000
```

But beta's **pre-upgrade** incarnation used an ephemeral counter that had climbed to at least **15915** over a long uptime, and gamma — which has not restarted in 33 h — still holds `max_seq = 15915` for beta.

So the new durable counter resumed roughly **5900 below** the high-water its peers already remembered. Every sequence beta emits in that band is a genuine Bloom hit (gamma really did see those values from old-beta) and is rejected as a replay.

Rejected sequences are climbing — 10919 → 10932 → 12974 — so this self-heals only when beta passes 15915: about **3.3 hours** at the observed ~15 msg/min. Each rejection is a severity-1.0 misbehaviour event in the meantime.

## Why it matters beyond this instance

This is the same *class* as the #2514 migration hazard: a fix changes the meaning of sequence state, but state written or remembered under the old regime is not healed. Two known instances:

- **Sender side (this issue):** the durable counter starts below what peers recorded from the ephemeral one.
- **Receiver side (#2514):** a node upgrading while holding a `+1000`-inflated persisted `max_seq` carries that inflated value forward as its new floor. Alpha escaped this only because `cleanup()` happened to delete and rebuild its entries; gamma and delta have not had that wipe.

## Suggested direction (not a decision)

Options worth weighing, none implemented:

1. On first start with a durable counter and no prior durable key, seed the reservation above a conservative estimate of any prior ephemeral high-water rather than from scratch.
[...]

Issue #2514, opened 2026-08-03
fix(net): a restarted receiver bans healthy peers for up to 1000 messages (replay safety gap)


fahertym:
Found while running the #2504 restart acceptance test on merged main `045fa9e9` (which carries #2505, #2510/#2511 and #2506). **This is the remaining blocker for #2504** — #2506 is fixed and proven, and removing it exposed this.

## What

`ReplayGuard::load_and_apply_safety_gap` (`icn-net/src/replay_guard.rs:196`) applies `RESTART_SAFETY_GAP` to **every peer's** stored high-water mark when the **receiving** node restarts:

```rust
let safe_max_seq = entry.max_seq.saturating_add(RESTART_SAFETY_GAP); // 1_000
window.max_seq = safe_max_seq;
window.floor_seq = safe_max_seq;   // reject ALL sequences at or below this
```

The floor itself is sound — after restart the bloom filter is empty, so a floor at the stored high-water is the only way to reject replays of pre-restart traffic. The **+1000** on top is not: it places the floor a thousand messages *above* anything the peer has actually sent.

A sender that did **not** restart has no reason to jump forward by 1000. So a restarted receiver rejects up to 1000 messages of perfectly legitimate traffic from a healthy peer — and escalates every one of them to `severity=1.0`, through quarantine, to a **ban**.

## Live evidence

Alpha and beta, both on `045fa9e9`. Alpha restarted at `00:18:35Z`:

```
00:18:52Z  Replay detected from <beta did>: sequence 10017 already seen (floor: 11011)
00:22:22Z  Replay detected from <beta did>: sequence 10070 already seen (floor: 11011)
```

Alpha's stored high-water for beta was **10011**; +1000 put its floor at **11011**. Beta — which is healthy and was not restarted by this action — is emitting from 10017 upward and is rejected on every message.

| | |
|---|---|
| alpha bans of beta, first 4 min | **34** |
| beta bans of alpha (symmetric) | **36** |
| alpha bans of *itself* | **0** (#2506 fixed) |
| beta bans of itself | **0** |
| beta sequence advance | 10017 → 10070 in 3m30s (~15/min) |
| sequences still to climb | 941 |
| **projected recovery** | **~63 minutes** |

Both directions are affected simultaneously, because each node applies the gap to the other.

## Why this was not visible before

[...]

Issue #2504, opened 2026-08-02
fix(security): a node restart can get it permanently banned by its peers


fahertym:
## Summary

Restarting a node can get it **permanently banned by its peers**, isolating it from the federation until the *banning* peer restarts. Observed live on the four-node K3s federation.

## Observed sequence

Alpha was rolled to a new image at `01:27:17Z`. At `01:27:48Z` — 31 seconds later — beta logged:

```
BANNING peer did:icn:zHdQuwTT... for severe violation   (remote_addr = alpha)
```

Fourteen minutes on, alpha had received **zero** gossip digests (each control received ~230 in the same period) and **zero** inbound connections from any real peer — 359 inbound, all from its own address. Alpha → beta still worked (beta logged 132 inbound from alpha); beta → alpha did not.

| node | received/s | digests/s | throttle % |
|---|---|---|---|
| alpha (restarted) | 0.163 | **0** | 16.9 % |
| beta | 0.648 | 0.270 | 2.1 % |
| gamma | 0.570 | 0.229 | 3.6 % |
| delta | 0.645 | 0.270 | 1.4 % |

## Mechanism

Reconnection after restart produces a burst. The peer's token bucket rejects part of it, which `icn-security::misbehavior` counts as violations. Past a threshold the peer is quarantined, then banned:

```rust
/// Ban a peer permanently
fn ban_peer(&mut self, did: &Did) {
    warn!("BANNING peer {} for severe violation", did);
    self.banned.insert(did.clone(), SystemTime::now());
    ...
}
```

`banned` is an in-memory map with **no expiry**. `is_banned` is a plain `contains_key`. A timestamp is stored but never consulted for release. So the ban lifts only when the banning process restarts.

## This is not caused by the rate-limit tier work

Worth stating explicitly, since it surfaced during the #2490/#2496 canary:

- The banning node (beta) runs the **old** image and applies its **own** old limits.
- Peer banning predates the change: beta banned gamma 3x at `18:00Z` the previous day, old image, old burst.
- The restarted node's own rate limiting was **not** implicated — every message it dropped was from its own DID (self-connection), none from any peer.

## Why it matters

Restarting a node is routine — deployments, rollouts, evictions, OOM kills. If routine restarts can earn a permanent ban from peers, then:

- rolling upgrades degrade the mesh cumulatively, one node at a time;
- **rollback does not recover** the node, because rollback is itself a restart that re-triggers the ban;
[...]

PR #2518, opened 2026-08-03
docs(architecture): record the #2514 live proof and the migration invariant

fahertym:
## What

Documentation only. Records the alpha-only live acceptance evidence for the merged #2514 fix (`e3c14c4d`), and adds a migration section stating an invariant this investigation surfaced.

## Live proof

Merged image deployed to alpha only. Beta ran continuously throughout (`restarts=0`, up since `00:17:15Z`); gamma and delta untouched. Three alpha restarts — one migration deploy and two deliberate receiver-only restarts — **zero replay rejections in every case**.

```
restart #1  06:51:58.795  command
            06:52:13.206  replay state loaded (no safety_gap field - pre-fix code gone)
            06:52:13.260  connected to beta            (+54 ms)
            ~06:52:15     digests flowing              recovery < 25 s

restart #2  06:56:59.582  command
            06:57:12.333  replay state loaded
            06:57:12.396  connected to beta            (+63 ms)
            06:58:13      6 digests received           no compounding
```

Against pre-fix behaviour on the same pair — 566 rejections, 566 severity-1.0 events, 566 bans, zero digests for a full hour — recovery is now bounded by connection setup rather than sender sequence burn or `cleanup()` timeout. Restart #2 is the compounding control: the old code added another `+1000` per restart, and the boundary did not move at all.

## The migration invariant

The more durable finding, and the reason this is a doc change rather than a note on the issue:

> Correct steady-state protocol semantics and safe migration from previously persisted or distributed semantics are **separate invariants**. Proving the first does not prove the second.

#2514's invariant is *restore exactly what was accepted*. That is faithful — including to a number written under earlier semantics that is already wrong. Two live instances are recorded: #2517 (sender's durable counter resumes below the ephemeral high-water peers recorded) and #2514's own upgrade case (a pre-fix inflated value is restored precisely). Both currently resolve only by peer-window timeout, which is the property #2504 exists to eliminate.

No bridge is designed here — that is #2517.

## Risk

None to running code. Registry host address omitted per the address-scrub rule; image identified by digest only.

Refs #2514, #2504, #2517

### DIFF
--- a/docs/architecture/replay-state-restart-invariants.md
+++ b/docs/architecture/replay-state-restart-invariants.md
@@ -258,7 +258,39 @@ is a prerequisite for signed-envelope freshness**, and therefore for replay-stat
   discovered.
 - **Failed `put` or `flush`.** Fail closed, per §6.
 
-## 9. Connection lifetime vs replay lifetime
+## 9. Migration: these invariants describe state created under the corrected protocol
+
+Every invariant above governs replay state **created by this code**. None of them repair state that
+was written, or is remembered by a peer, under earlier sequence semantics. That is a separate
+invariant and it is not yet satisfied.
+
+The distinction is sharp enough to be worth stating as a rule:
+
+> **Correct steady-state protocol semantics and safe migration from previously persisted or
+> distributed semantics are separate invariants.** Establishing the first does not establish the
+> second, and a faithful restore of a legacy value is still wrong.
+
+Two known instances, both live-observed:
+
+- **Sender side (#2517).** `SigningSequenceCounter` initialises its durable counter without any
+  bridge from the ephemeral counter it replaced, so a first upgrade can resume *below* the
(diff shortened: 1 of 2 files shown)


fahertym:
Both review points fixed in `e114322a`.

1. **Invariant wording.** The investigation doc paraphrased ("Proving the first does not prove the second") where `replay-state-restart-invariants.md` says "Establishing the first does not establish the second, and a faithful restore of a legacy value is still wrong." It now quotes the invariants doc verbatim and names it as the owner, so the pair cannot drift into two wordings of one invariant.

2. **"Three alpha restarts" vs two timing blocks.** Three did happen; the third is the deploy that crossed the code boundary, so it has no comparable pre-fix timing line. The prose now says that explicitly instead of leaving prose and evidence in apparent disagreement.

PR #2520, opened 2026-08-03
fix(net): bind Hello DID claims to the current connection certificate

fahertym:
## What

The Hello handler verifies that the claimed DID signed a TLS certificate hash, but never
checks that hash against the certificate presented by the connection carrying the Hello.
This wires that check into the production path, before any state keyed on the claimed DID
is stored.

Handled privately under **GHSA-c22h-qr8j-g639**. Details of impact and exploitation are in
the advisory, not here.

## Why

`icn-identity::verify_binding_info` — which performs the current-certificate comparison —
has existed and been unit-tested since 2025-11-13. It has had no production caller. The
Hello path called `verify_did_matches_binding`, which proves the binding belongs to the
sender but says nothing about the connection carrying it.

The strict check was originally wired in as `if let Some(peer_cert) = peer_identity()`.
When client auth was made optional to stop connections breaking, that condition became
sometimes-false and the check was removed rather than made fail-closed.

## How

- `handle_hello` verifies the binding against the current connection's certificate before
  storing anything derived from the claimed DID. Both connection directions funnel through
  this one handler, so a single call site covers inbound and outbound.
- Connections presenting **no** peer certificate are refused. Client auth is offered but
  not mandatory, so treating absence as "cannot check, therefore allow" would let a peer
  opt out of authentication by omission. Measured: an anonymous inbound peer completes the
  handshake with `peer_identity() == None`.
- `NetworkMessage::verify_hello` now also checks the binding's DID against the sender; it
  compared the certificate hash but never whose binding it was.
- New `icn_network_hello_binding_rejected_total{reason}`. Labelled by failure class only,
  never by DID — the peer is unauthenticated, so the DID is a name it chose.
- No misbehaviour scoring on binding failure, for the same reason: it would let anyone
  degrade the reputation of a peer they do not control.

## Risk

Legitimate nodes are unaffected: `SessionManager::start` builds both TLS configs from
`identity_bundle.tls_cert()`, so the certificate presented in either direction is exactly
the one hashed into `BindingInfo`. Verified by a positive control that passes both before
and after the change.

Rotation is safe — `IdentityBundle` derives the binding from the certificate it holds, so
the two cannot drift apart.

The behaviour change is that a peer presenting no client certificate can no longer
establish DID-attributed session state. That is the intended fix, and no ICN client does
this.

## Test plan

New `crates/icn-net/tests/hello_current_cert_binding.rs` — real QUIC, real TLS config,
real dispatch loop, real handler; exercises the composition path rather than the helper.

- 3 failed / 3 passed before the change; 6 passed / 0 failed after.
- Positive control (matching certificate accepted) passes in both states, so the fix is
  not "reject everything".
- Controls for wrong DID, tampered signature, absent certificate, and non-corruption of an
  already-established peer entry.
- A composition guard fails if a future path picks up the weak verifier on its own.

Mutation testing — 4 mutations applied, 4 killed, 0 survivors:

| Mutation | Killed by |
|---|---|
| skip current-cert comparison | `hello_replayed_onto_a_different_current_cert_is_rejected`, `forged_hello_does_not_corrupt_established_peer_state` |
| fail open when certificate absent | `hello_without_any_peer_certificate_is_rejected` |
| compare against stored cert not current | `hello_with_matching_current_cert_is_accepted`, `forged_hello_does_not_corrupt_established_peer_state` |
| drop binding-DID vs sender check | `hello_claiming_another_did_with_own_valid_binding_is_rejected` |

Local: `icn-net`, `icn-identity`, `icn-obs` full suites green; fmt, clippy `-D warnings`,
meaning firewall, and doc control green.

Refs GHSA-c22h-qr8j-g639


icn/crates/icn-obs/src/metrics/network.rs
```diff
@@ -272,6 +277,20 @@ pub fn version_negotiation_failure_inc(reason: &str) {
     .increment(1);
 }
 
+/// A Hello was refused because the claimed DID could not be bound to the certificate
+/// presented by the current QUIC connection.
+///
+/// The label is a coarse failure class, never a DID: an unauthenticated party chooses
+/// the DID it claims, so labelling by DID would let anyone forge series for a peer they
+/// do not control.
```
fahertym, line 285: Both addressed in `ac798c3d`.

Metric: verified — `init_descriptions()` has 34 `describe_counter!` entries including `icn_network_version_negotiation_failures_total`, the one I patterned this on. Registered it with a help line noting the `reason` label is coarse and never a DID.

Doc index: added to `docs/INDEX.md` next to the sibling network invariant docs.

I did not touch `DOCUMENT_REGISTRY.md`. It is generated, its header is stamped `Last Reviewed: 2026-07-28`, and it does not list `network-identity-self-exclusion.md` or `replay-state-restart-invariants.md` either — so hand-adding one entry would introduce drift rather than remove it. If it should track architecture docs, regenerating it is a separate change.

Issue #2564, opened 2026-08-09
fix(gateway): IPv4-mapped addresses bypass the is_private_ip SSRF blocklist


fahertym:
Found while sweeping for the same root cause as #2563 (and #2557's `SourceKey` fix). Filed
separately because it is a different crate, a different policy, and a security control rather than
a classification helper — it deserves its own review.

## Defect

`icn/crates/icn-gateway/src/api/listings.rs`, `fn is_private_ip`, branches on `IpAddr`. The
`IpAddr::V6` arm checks only `is_loopback()`, `is_unspecified()`, ULA (`fc00::/7`) and link-local
(`fe80::/10`). It never unmaps IPv4-mapped addresses, and Rust does not unmap implicitly.

`validate_photo_url` parses the URL host as a literal IP (`host.parse::<IpAddr>()`) and rejects it
when `is_private_ip` is true. So every IPv4 range the blocklist covers can be expressed in mapped
form and pass:

```
https://[::ffff:127.0.0.1]/           -> V6 arm; is_loopback() is false for mapped 127/8 -> allowed
https://[::ffff:10.0.0.1]/            -> allowed
https://[::ffff:192.168.1.1]/         -> allowed
https://[::ffff:169.254.169.254]/     -> allowed   (cloud instance metadata)
https://[::ffff:100.64.0.1]/          -> allowed   (CGNAT)
```

The string prefilter above it does not catch these either — it compares against the literal
`"127.0.0.1"`, `"localhost"` and reserved TLDs, none of which match `::ffff:127.0.0.1`.

`Ipv6Addr::is_loopback()` is true only for `::1`; it is never true for a mapped `127/8` address.
That is the same gap #2563 fixes in `peer_exchange::endpoint_kind_for_addr` and that #2557 fixed in
`SourceKey::from_addr`.

## Severity — stated carefully

The function's own doc comment describes it as anti-SSRF: "preventing SSRF attacks regardless of
URL length or format." As written, that control is bypassable for the entire IPv4 blocklist.

However, **no server-side fetch of these URLs exists today**. The only outbound HTTP client in the
gateway is `fcm_client.rs` (Firebase push), which does not consume listing photo URLs. So this is a
**bypassable defense-in-depth control, not a demonstrated live SSRF**.

It still matters, because:

- the control is documented as SSRF prevention, so future code may rely on it
- the stored URL is served to clients, and a client that fetches it resolves the private address on
  *its own* network
[...]

Issue #2563, opened 2026-08-09
fix(net): normalize IPv4-mapped addresses before peer endpoint classification


fahertym:
Found by inspection during the final merge gate for #2557 (squash `2ca06887`). Pre-existing on
main; not introduced by that PR. Same root cause as the `SourceKey` defect #2557 fixed, in a
different subsystem.

## Defect

`icn/crates/icn-net/src/handlers/peer_exchange.rs`, `fn endpoint_kind_for_addr`, branches on
`addr.ip()`. The `IpAddr::V6` arm recognises only `::1`, ULA (`fc00::/7`) and link-local
(`fe80::/10`) as `Local`. It never unmaps IPv4-mapped addresses.

The default `listen_addr` is `[::]:7777` (`icn-core/src/config/mod.rs`), production binds it with a
plain `std::net::UdpSocket::bind` (`icn-net/src/session.rs`) and nothing sets `IPV6_V6ONLY`. On a
dual-stack socket an IPv4 peer is reported as `::ffff:a.b.c.d`, and Rust does not unmap implicitly.
So:

```
::ffff:192.168.1.5  -> v6 arm -> not ULA, not link-local -> Public   (should be Local)
::ffff:127.0.0.1    -> v6 arm -> Ipv6Addr::is_loopback() is false    -> Public   (should be Local)
```

That is precisely the outcome the function's own doc comment says it prevents: "This prevents
misclassifying LAN peers as publicly reachable when the observed remote address is non-routable."

## Actual consequence, from the call sites

One production call site: the peer-exchange **response** path
(`peer_exchange.rs`), which labels each connected peer's `remote_address()` before advertising it.

- **No new information exposure.** There is no kind-based filtering before advertisement — the
  address is sent whatever its label. What is wrong is the label, not the disclosure. (Disclosure
  policy is #2535's lane.)
- **Dial tiering is wrong.** `supervisor::init_network`'s Announce path builds a priority-ordered
  list `Local -> Public -> Relay` with `MAX_ENDPOINTS_PER_KIND = 3`, so a mapped LAN address is
  tried in the wrong tier, later than it should be, and consumes a slot in the Public cap rather
  than the Local one.
- **Wrong dial timeout and metric.** `supervisor::nat_dial` splits candidates by kind into separate
  Happy Eyeballs races with `local_dial_timeout_ms` vs `public_dial_timeout_ms`, and reports
[...]

Issue #2556, opened 2026-08-08
refactor(net): make authenticated peer state structurally writable only after verified Hello


fahertym:
## Current production state

**This is not a demonstrated authentication bypass, and nothing here is exploitable today.**

`ConnectionContext::record_authenticated_peer` is called from exactly one place — `handlers/hello.rs`, below all three DID↔TLS checks — and the `authenticated_peer` field has exactly one direct writer, inside that setter. Independent audit during #2554 found no gap between any failing check and its `return`, and no second path that establishes authenticated state. Production ordering is sound.

This issue is about **future** structural regressions, not a present defect.

## Why this exists

PR #2555 asserts the ordering *inside* `handle_hello`. Adversarial review established that direct-handler tests structurally cannot see a regression introduced **one frame above** the handler: a failed check returns `Err`, the stream loop ends, and the transport is destroyed — so "peer entry absent" and "connection closed" hold whether or not a bogus identity was bound first.

The following were each **demonstrated during review** to authenticate a sender-chosen DID before verification while leaving the entire relevant test surface green:

| shape | proven |
|---|---|
| dispatcher calls the setter above `ctx.handle_hello(...)` | all six behavioural tests + 5 existing suites green |
| direct `authenticated_peer` write in another handler | neither runtime observable moves |
| second helper in the authorised file, reached from the dispatcher | green |
| second direct writer beside the setter | green; also never drops the admission guard |
| alternate lock accessors (`try_write`, `write_owned`, …) | green |
| function-item alias of the setter | green |
| write extracted into a helper the setter delegates to | green |

## Why a source scanner is the wrong tool

A whole-crate lexical guard was built and iterated across five review rounds. Each round closed one *spelling* and the next escape appeared:

1. accessor allow-list → missed `try_write`
2. `split_once("//")` → a `//` inside a string literal hid the rest of the line
3. `.X(` / `::X(` needles → missed a function-item alias
4. cardinality → cannot see a function boundary (helper extraction)
5. `"fn "` substring → missed `fn(`, the function-pointer spelling

Every one of these was a **valid** reviewer finding. The pattern is the conclusion: *source spelling is the wrong enforcement boundary*. A text scan cannot express "only this function may mutate this state" — that is a type- and module-level property. The scanner was removed from #2555 rather than patched a sixth time.

## Desired property

> Code outside the verified authentication transition should be structurally unable to mutate authenticated peer state at all — enforced by `rustc`, not by a test-time text scan.

[...]

Issue #2550, opened 2026-08-07
design(net): nothing bounds how many physical connections one source may hold once authenticated


fahertym:
Found while reviewing #2547's invariant. Filed for ownership, not folded into #2547 — that issue is about *anonymous* admission, and this is about what happens after a peer identifies itself.

## The escape hatch

#2547 releases a connection's pre-authentication slot at the authenticated Hello, deliberately: the bound is on connections a source holds *while refusing to say who it is*, which is what keeps it off legitimate NAT'd peers. That release is reachable cheaply:

```text
generate an ephemeral DID            -> Ed25519 keygen, essentially free
connect, send a valid Hello          -> DID-TLS binding over this connection's cert (#2520)
                                     -> pre-auth slot released
hold the connection open
repeat with a fresh DID
```

Each iteration is *cryptographically* valid: the binding genuinely proves the peer holds that DID's key. Nothing about #2520 is bypassed. But DIDs are unbounded and free to mint, so "authenticated" is not by itself a scarce resource.

## What current session-map semantics do and do not bound

`SessionManager::connections` is a plain `HashMap<String, quinn::Connection>` with **no cardinality cap** (`icn-net/src/session.rs`). Its eviction logic (#2530) matches on `stable_id` and only ever drops an *alias* of the connection in hand — it explicitly must not drop a different connection to the same peer, and there is a test pinning that. So:

- **bounded:** aliases of one connection; keys superseded by an authenticated DID.
- **not bounded:** the number of distinct DIDs, and therefore the number of simultaneous authenticated connections one source may hold.

Each retains a handler task, a `ConnectionContext`, quinn connection state, and advertised flow-control credit. The server's own `keep_alive_interval(30s)` against `max_idle_timeout(60s)` means these do not idle out either.

## What already limits the damage

Post-authentication traffic is charged to the authenticated DID under #2490/#2491, and under `EnforcementMode::RequirePersonhood` a DID with no anchor is denied outright. So an ephemeral-DID fleet gets connections, not throughput. The open question is whether holding the connections themselves is worth bounding.

## Not a defect in #2547

#2547's invariant is about anonymous connections, and it holds. This is the adjacent question of whether authenticated multiplicity needs its own bound — plausibly a per-source cap on *total* connections regardless of phase, or an anchor-derived one. Needs a design decision before any implementation.

Refs #2547, #2491, #2520, #2530

PR #2568, opened 2026-08-10
fix(gateway): trust forwarded client IP only from trusted proxies

### LINKED ISSUE #2567: fix(gateway): X-Forwarded-For is trusted unconditionally for rate-limit and audit identity
Found during the adversarial review of #2566 (#2564, gateway literal-IP validation). Filed
separately: different function, different trust boundary, and it should not ride on a URL-validation
fix. **No change for this was made in #2566.**

## Defect

Three separate copies of `get_client_ip` in `icn-gateway` return the leftmost `X-Forwarded-For`
element with **no trusted-proxy gate at all**:

- `icn/crates/icn-gateway/src/api/auth.rs:16`
- `icn/crates/icn-gateway/src/api/listings.rs:32`
- `icn/crates/icn-gateway/src/api/sessions.rs:48`

All three are shaped:

```rust
if let Some(forwarded) = req.headers().get("x-forwarded-for") {
    if let Ok(forwarded_str) = forwarded.to_str() {
        if let Some(client_ip) = forwarded_str.split(',').next() {
            return client_ip.trim().to_string();   // <- caller-supplied, unvalidated
        }
    }
}
// falls back to req.peer_addr() only when the header is absent
```

The header is trusted unconditionally, and `.split(',').next()` selects the **leftmost** element —
the part a client controls end-to-end.

## Why this is not merely a deployment assumption

The repository's own proxy configuration encodes the appending behaviour that makes the leftmost
element caller-controlled:

- `deploy/config/nginx.conf:20` — `proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;`
  for `location /v1` → `proxy_pass http://icnd:8080`. `$proxy_add_x_forwarded_for` **appends** the
  peer address to whatever the client sent; it does not replace it. A client sending
  `X-Forwarded-For: <chosen-value>` causes the gateway to receive
  `<chosen-value>, <real-peer>`, and `get_client_ip` takes `<chosen-value>`.
- `deploy/appliance/lan/nginx-icn-rehearsal-https.conf.in:46` and the `-http` sibling proxy
  `location /v1/` → `127.0.0.1:8080` setting only `Host`. No `X-Forwarded-For` handling at all, so
  a client-supplied header reaches the gateway verbatim.
- Direct access to the gateway (no proxy in front) is a shipped posture:
  `deploy/appliance/systemd/icnd.service.d/20-demo-profile.conf:38` (`--gateway-bind 0.0.0.0:8080`),
  `deploy/k8s/reference/icn-beta-deployment.yaml:43`, `deploy/k8s/multi-node/scripts/deploy-coop.sh:134`.

No middleware strips or rewrites the header before handlers run — `server.rs` wraps metrics, CORS,
security headers, tracing and compression only.

## What consumes the value

`IpRateLimiter::check_rate_limit(&client_ip)`, on endpoints that are unauthenticated by design:

| call site | route |
| --- | --- |
| `api/auth.rs:47` | `POST /v1/auth/challenge` |
| `api/auth.rs:77` | `POST /v1/auth/verify` |
| `api/sessions.rs:167` | `POST /v1/sessions` |
| `api/sessions.rs:214` | `GET /v1/sessions/{session_id}` |
| `api/listings.rs:949` | `POST /v1/listings/{id}/interest` |

and audit attribution: `AuditLogger::log_auth_attempt(..., &client_ip)` (`api/auth.rs:122,129`) and
`AuditLogger::log_invalid_scope_request(..., &client_ip)` (`api/auth.rs:92`).

Consequences, stated narrowly:

1. **The IP rate limit is a bound on a header, not on a source.** A caller that varies the header
   per request gets a fresh bucket each time, so the limiter does not bound what it is documented to
   bound (`api/auth.rs:45` — "SECURITY: IP-based rate limiting for unauthenticated endpoint").
2. **Audit records can be attributed to an address the caller chose.**

The ceiling itself is real, but the key it is applied to is free to mint, so the bound constrains a
behaviour rather than a resource.

## The codebase already contains the correct pattern

`api/sessions.rs:68`, `get_gateway_url`, roughly twenty lines below the ungated `get_client_ip` in
the same file, gates `X-Forwarded-*` on a `TRUSTED_PROXY_IPS` allowlist and logs a warning when the
peer is untrusted:

```rust
// Only trust X-Forwarded-* headers if from a trusted proxy
if is_trusted_proxy { ... } else if req.headers().contains_key("x-forwarded-for") { warn!(...) }
```

So this is an inconsistency within one file, not a missing concept.

## Severity

**Security control bypass — the rate limit and the audit identity on unauthenticated endpoints do
not bind an untrusted caller.** Not a remote code execution or an authentication bypass: a forged
`X-Forwarded-For` does not grant credentials, and JWT-authenticated routes are unaffected. Not
verified against a live deployment; the evidence above is code and configuration in this repository.

## Suggested direction (not prescriptive)

One shared, gated accessor rather than three copies:

- reuse the `TRUSTED_PROXY_IPS` allowlist `get_gateway_url` already reads, defaulting to loopback;
- use `req.peer_addr()` unless the peer is on the allowlist;
- when it is, take the rightmost untrusted element rather than the leftmost, or a hop count.

Two details worth deciding explicitly during that work:

- the existing allowlist is an **exact string** comparison against `peer_addr.ip().to_string()`
  (`api/sessions.rs:83`), so a dual-stack peer arriving as `::ffff:a.b.c.d` fails to match a
  `TRUSTED_PROXY_IPS` entry written in dotted form. That direction is fail-closed today (see #2564 /
  #2566 for why the mapped-address normalization there was deliberately kept local), but it is an
  operability trap once this list becomes load-bearing;
- `get_client_ip` returns a `String`, so the limiter is keyed on unparsed text. Parsing to `IpAddr`
  would also close the dual-stack key-splitting class fixed in #2557 / #2563.

## Scope note

Not fixed in #2566, which is confined to literal-IP URL validation in `validate_photo_url`. Refs
#2564, #2566, #2557, #2563.

fahertym:
## What

`X-Forwarded-For` was trusted unconditionally. Three separate copies of `get_client_ip` in
`icn-gateway` returned the **leftmost** element of the header with no check on who sent it, and
that value became the `IpRateLimiter` key and the audit client IP.

This replaces all three with one gated derivation.

## The authority invariant

> Forwarded client identity is an untrusted request claim unless the immediate transport peer is
> explicitly trusted to assert forwarding metadata. If the immediate peer is not trusted, the
> socket peer IP is the client IP.

## Why this is not "take the rightmost XFF value"

Picking the rightmost element is a different bug, not a fix — it still reads a caller-supplied
header without establishing that anyone trustworthy wrote it, and it breaks the moment there are
two proxies. The implemented algorithm is a **trust-gated traversal**:

1. Resolve the immediate transport peer (`req.peer_addr()`).
2. If that peer is **not** on the trusted-proxy allowlist: ignore `X-Forwarded-For` entirely, the
   socket peer is the client. Done.
3. If it **is** trusted: walk the chain **right to left** — from the transport-peer side inward —
   **skipping hops that are themselves trusted proxies**. The first hop no trusted proxy vouched
   for is the client.
4. If every hop is a trusted proxy, the furthest-upstream element is as far as the trusted chain
   reaches. An absent, empty, or unparseable chain falls back to the socket peer.

Step 3 is what makes a multi-proxy chain resolve correctly, and it is why anything a caller
prepends can never win: the caller's bytes sit to the *left* of whatever the trusted proxy
appended, so the right-to-left walk reaches the proxy-asserted hop first.

## Behaviour table

| immediate peer | `X-Forwarded-For` | effective client IP |
| --- | --- | --- |
| untrusted `198.51.100.42` | *(absent)* | `198.51.100.42` |
| untrusted `198.51.100.42` | `203.0.113.7` | `198.51.100.42` — header ignored |
| untrusted `198.51.100.42` | `203.0.113.7, 198.51.100.42` | `198.51.100.42` — header ignored |
| trusted `127.0.0.1` | `203.0.113.7` | `203.0.113.7` — proxy-asserted |
| trusted `127.0.0.1` | `203.0.113.7, 198.51.100.42` | `198.51.100.42` — prepended claim discarded |
| trusted `127.0.0.1` | `198.51.100.42, 127.0.0.1, ::1` | `198.51.100.42` — trusted hops skipped |
| trusted `127.0.0.1` | `not-an-ip` | `127.0.0.1` — falls back to peer |
| *(no transport peer)* | anything | `unknown` |

## Pre-fix reproduction — route level, not a parsing helper

The reproduction drives the real `POST /v1/auth/challenge` route with the real `IpRateLimiter`,
and observes limiter *identity* behaviourally: drain the bucket for one socket peer, then check
whether a caller-supplied header can mint a fresh one. A test that called the extraction helper
directly would prove what a comma parser returns, not what the limiter is keyed on.

Against the pre-fix code, 3 of the 4 route-level assertions went red:

| socket peer | `X-Forwarded-For` | expected after bucket exhaustion | observed |
| --- | --- | --- | --- |
| `198.51.100.42` | `203.0.113.1` | 429 | **200** |
| `198.51.100.42` | `203.0.113.1, 198.51.100.42` | 429 | **200** |
| `127.0.0.1` | `203.0.113.1, 198.51.100.42` | 429 | **200** |

Each drain step asserts it actually observed a 429 first, so a later "still limited" assertion
cannot pass vacuously.

The fourth test — distinct forwarded clients behind a trusted proxy must keep distinct buckets —
passes both before and after. It is deliberately included as a guard against over-correcting into
"always key on the socket peer", which would collapse every client behind a proxy into one bucket.

## Consequence 1: the IP rate limit bounded a header, not a source

`IpRateLimiter::check_rate_limit(&client_ip)` on:

**Public / unauthenticated by design (4):**

| call site | route |
| --- | --- |
| `api/auth.rs` | `POST /v1/auth/challenge` |
| `api/auth.rs` | `POST /v1/auth/verify` |
| `api/sessions.rs` | `POST /v1/sessions` |
| `api/sessions.rs` | `GET /v1/sessions/{session_id}` |

**Authenticated (1):**

| call site | route | protection |
| --- | --- | --- |
| `api/listings.rs` | `POST /v1/listings/{id}/interest` | `auth.clone()` on the `/listings` scope **plus** `require_scope("coop:write")` |

This corrects the issue report, which listed all five as "unauthenticated by design". Only the
first four are reachable without a credential; the listings route is defence-in-depth IP limiting
layered under JWT auth.

Varying the header per request minted a fresh bucket each time, so the ceiling constrained a
behaviour rather than a resource.

## Consequence 2: caller-controlled network-origin attribution in audit records

`api/auth.rs` derives one `client_ip` and passes it to both the limiter and the audit sink:
`AuditLogger::log_auth_attempt(...)` and `AuditLogger::log_invalid_scope_request(...)`. A caller
could therefore choose the **network-origin / client-IP field** recorded against an authentication
attempt.

Scope of that claim, stated narrowly: this is attribution of the origin address only. It is **not**
DID impersonation, **not** authenticated-user impersonation, and **not** an authorization bypass —
the DID and its signature verification are untouched by the header, and a forged `X-Forwarded-For`
grants no credential.

Proved rather than asserted: the tests install a thread-local `tracing` subscriber, capture the real
`auth_attempt` event, and assert on its `ip` field. Both were re-run with the old `get_client_ip`
temporarily restored and went red on their own assertions, so they are not vacuous green.

## Changes

**One canonical derivation** — new `icn/crates/icn-gateway/src/client_ip.rs` (`pub(crate)`),
exposing `client_ip`, `client_ip_key`, and `is_trusted_proxy`.

**Three former copies migrated** to call it:

- `icn/crates/icn-gateway/src/api/auth.rs`
- `icn/crates/icn-gateway/src/api/listings.rs`
- `icn/crates/icn-gateway/src/api/sessions.rs`

**Trust model unified.** `api/sessions.rs::get_gateway_url` already gated `X-Forwarded-*` on
`TRUSTED_PROXY_IPS`, about twenty lines below the ungated `get_client_ip` in the same file. It had
its own inline predicate doing an exact **string** comparison against `peer_addr.ip().to_string()`.
It now calls `client_ip::is_trusted_proxy`, so the gateway has one trusted-proxy model instead of
two that could drift.

**Dual-stack / IPv4-mapped handling.** Allowlist entries and peer addresses are compared as parsed
`IpAddr` after unwrapping IPv4-mapped IPv6 (`::ffff:a.b.c.d`) via `to_ipv4_mapped` — strictly
`to_ipv4_mapped`, not `to_ipv4`, so deprecated IPv4-*compatible* (`::a.b.c.d`) addresses are not
folded in.

The stated goal is narrow: prevent one logical address from occupying two different identities
purely because it appeared once as native IPv4 and once as strict IPv4-mapped IPv6 — the key-split
class of #2557 / #2563 — and stop a dual-stack peer from silently failing to match a dotted
allowlist entry. Reviewers should attack specifically whether mapped-form equivalence is correct for
**trusted-proxy authority** as well as for **client limiter identity**; those are two different
questions and this PR answers both the same way.

**Configuration semantics** (`TRUSTED_PROXY_IPS`, unchanged variable, now load-bearing for more):
exact addresses, comma-separated, no CIDR. Unset means loopback only (`127.0.0.1`, `::1`) —
preserving the prior default. Set-and-empty means trust nobody. Unparseable entries are skipped;
they could not match a kernel-supplied peer address anyway.

## Appliance nginx: a trusted proxy must *assert*, not relay

Found during implementation, not stated in the issue, and kept here because it is the same
authority boundary rather than a separate deployment concern.

`deploy/appliance/lan/nginx-icn-rehearsal-http.conf.in` and its `-https` sibling proxy `/v1/` to
`127.0.0.1:8080` while setting only `Host`. Two facts combine badly:

- the gateway sees a **loopback** transport peer, which is **trusted by default**;
- nginx forwarded the client's `X-Forwarded-For` **verbatim**, asserting nothing of its own.

So on that posture a correct application-side gate is still bypassed. The attacker's claim arrives
through a peer the application legitimately trusts and comes out the other side authoritative.

The general failure: **a trusted-proxy gate authenticates the sender of the header, not the
provenance of its contents.** Only the proxy can distinguish the two, by writing the address it
actually observed into the chain. A trusted-but-silent proxy is worse than an untrusted one,
because the untrusted one fails closed.

Both templates now set:

```nginx
proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
```

`$proxy_add_x_forwarded_for` appends the address nginx actually saw, so the right-to-left walk
reaches it first and discards anything prepended. `deploy/config/nginx.conf` already did this; the
review pass added an operator comment to it (see below) but left its behaviour untouched.

## Deployment requirement — reverse proxy / ingress

> **Deployments behind a reverse proxy or ingress must explicitly configure that proxy in
> `trusted_proxy_ips`.** Otherwise forwarded identity is intentionally ignored and every client
> behind the proxy shares the proxy's transport IP as its per-IP limiting identity.

That rule is general. **It does not describe the shipped Kubernetes topology**, and the review
settled that question rather than leaving it open:

- `deploy/k8s/` ships **no `Ingress` object**. The gateway is reached over **NodePort 30080**
  (`services.yaml`), i.e. directly.
- Both `configMapKeyRef`s (`gateway_base_url`, `trusted_proxy_ips`) are `optional: true`, so the
  commented-out key cannot break pod startup.
- With `externalTrafficPolicy` unset (default `Cluster`), kube-proxy SNATs the client address to
  the node **regardless of this PR**, so per-client attribution is already collapsed there and no
  proxy is setting `X-Forwarded-For` to recover it.

So leaving `trusted_proxy_ips` commented out is **correct for what is actually shipped**, and this
PR does not make the default Kubernetes deployment worse. Classification: **NOT A DEFECT**.

The general rule still binds any operator who *adds* a reverse proxy or ingress: without an explicit
entry the gateway sees the proxy's transport IP, does not trust it, ignores `X-Forwarded-For`, and
keys every client behind it on that one address — against an `IpRateLimiter` of capacity 20, refill
2/s. That is **fail-closed** behaviour, not a spoofing bypass, but it will limit real users. This PR
updates the configmap comment to make the requirement discoverable and **deliberately does not guess
a production ingress address**.

Explicitly not done: trusting all RFC1918, trusting a cluster CIDR, trusting `X-Forwarded-For` when
the allowlist is empty, or auto-detecting proxies.

## Confirmed not affected

`rate_limit.rs:549` builds the anonymous-DID limiter key from `connection_info().peer_addr()`, which
returns the socket peer only. The `X-Forwarded-For`-consulting accessor in actix-web is
`realip_remote_addr()`, and it has **zero callers** in this repository. That path is outside this
change and is not broadened into.

## Tests

New `icn/crates/icn-gateway/tests/forwarded_client_ip_trust.rs` — 6 tests, all through the real
route:

- `untrusted_peer_cannot_mint_buckets_with_forwarded_for`
- `untrusted_peer_cannot_mint_buckets_with_appended_chain` (the shipped nginx chain shape)
- `trusted_peer_ignores_caller_prepended_chain_element`
- `trusted_peer_still_separates_distinct_forwarded_clients` (over-correction guard)
- `audit_attribution::untrusted_peer_is_audited_by_its_socket_address`
- `audit_attribution::trusted_peer_audits_the_proxy_asserted_client`

Plus 10 unit tests in `client_ip.rs` covering the traversal, trusted-hop skipping, mapped-address
normalization, unparseable input, and the no-peer case. Those read the default (unset) policy only,
so they never mutate the process-global `TRUSTED_PROXY_IPS` and cannot race the rest of the suite.

**Red proofs.** Route-level: 3 of 4 red pre-fix, on the intended assertions (200 where 429 was
expected). Audit: both tests re-run against a temporarily restored vulnerable `get_client_ip`, both
red, then the revert was undone and re-verified.

## Local validation

```
cargo fmt --all --check                                              clean
cargo clippy -p icn-gateway --all-targets --all-features -D warnings exit 0
cargo test -p icn-gateway --all-features                             47 binaries, 0 failures
cargo test -p icn-gateway --test forwarded_client_ip_trust           6 passed, 0 failed
```

## Risk

Behaviour changes for any deployment that was relying on `X-Forwarded-For` from an untrusted peer —
by design, that is the bug. The failure direction is closed (identity collapses onto the transport
peer), not open. The over-correction guard test covers the legitimate proxied case. `get_gateway_url`
keeps its existing semantics; only its trust predicate moved.

## Non-goals

- Not a general reverse-proxy framework.
- No RFC 7239 `Forwarded` support, and no `X-Real-IP` support — neither is read today and neither is
  set by any proxy config in this repository; reading them would widen attacker-controlled input for
  no deployment benefit.
- No CIDR support in the allowlist; exact addresses only, as before.
- No DNS or IP-policy changes.
- No automatic Kubernetes ingress discovery, and no invented ingress address.
- No changes to the `connection_info().peer_addr()` anonymous-DID path.
- No work on #2549, and none on #2550 or #2556.

Closes #2567

Refs #2564, #2566, #2557, #2563

---

## Independent adversarial review pass (`f3339512`)

A fresh session re-derived the patch from `origin/main..HEAD` without inheriting the implementation
session's trust model, reproduced the pre-fix defect by isolated mutation, and ran an independent
43-case matrix through the real `POST /v1/auth/verify` audit path plus five targeted mutations.
Result: the authority invariant holds. One residual finding was fixed on this branch.

### Fixed: only the first `X-Forwarded-For` field-line was read

`client_ip` collected the chain with `HeaderMap::get`, which returns the **first** field-line only.
A trusted proxy that asserts its observation by **appending a separate header** rather than
rewriting the caller's leaves two field-lines on the wire — and the caller's is first. The
attacker's claim therefore ended up to the *right* of the proxy's assertion in the chain the
right-to-left walk actually saw, and won it.

Measured through the real audit path, trusted peer `127.0.0.1`, two headers
`X-Forwarded-For: 203.0.113.7` then `X-Forwarded-For: 198.51.100.42`:

| | effective client IP |
| --- | --- |
| before | `203.0.113.7` — caller's claim |
| after | `198.51.100.42` — proxy-asserted |

**Not reachable in any shipped topology**: both appliance templates and `deploy/config/nginx.conf`
use `proxy_set_header`, which emits a single merged field-line, and nginx already folds duplicate
client-sent `X-Forwarded-For` headers into one `$http_x_forwarded_for` value. It is fixed anyway
because it made the gate depend on a proxy's header-emission style rather than on the trust
decision. Per RFC 7230 3.2.2 repeated field-lines are equivalent to one comma-joined field-line, so
reading all of them reconstructs the true chain.

### Verified, not merely re-read

- **Untrusted peer**: header content is irrelevant across absent / single / multi / IPv4 / IPv6 /
  strict-mapped / malformed / empty / `unknown` / `host:port` / `[bracketed]` / 60k-comma inputs —
  all 11 produce the socket peer.
- **Strict mapped-v4 trust authority** is the correct narrow choice. `::ffff:127.0.0.1` inherits
  loopback authority; IPv4-*compatible* `::127.0.0.1`, NAT64 `64:ff9b::7f00:1`, and IPv4-translated
  `::ffff:0:7f00:1` all do **not**. That is exactly the `to_ipv4_mapped` / `to_ipv4` distinction.
- **Config parsing fails closed**: a CIDR typo, junk, or an empty value yields an empty allowlist —
  trust nobody, including loopback. `0.0.0.0` is a literal address, never a wildcard.
- **Multi-proxy** traversal stops at the first untrusted hop from the trusted edge; it does not walk
  past an untrusted intermediary to reach an earlier claim.
- **Chain work is bounded** by actix-http's 128 KiB request-head limit (`MAX_BUFFER_SIZE`), with
  `client_request_timeout(30s)` bounding the read. Unparseable entries allocate nothing.
- **Mutation controls**: reverting the peer gate, the walk direction, forwarded metadata entirely,
  strict-mapped normalization, and the header-merge each turn at least one test red; the pre-fix
  `get_client_ip` turns 5 of 6 integration tests red while every unit test stays green — which is
  precisely why the suite enters through the route.

### Also recorded (no behaviour change)

- A non-empty `TRUSTED_PROXY_IPS` **replaces** the loopback default rather than extending it, so an
  operator adding an ingress silently de-trusts an on-host proxy. The appliance nginx reaches the
  gateway over `127.0.0.1`, so this matters there. Documented in `client_ip.rs`.
- `deploy/config/nginx.conf` asserts the header correctly, but the gateway sees that container's
  compose-network address, not loopback — so it is untrusted by default and every client behind it
  now shares one bucket. Fail-closed, but it needed saying where the operator looks.

### Out of scope, filed as observations rather than acted on

- `api/sessions.rs` `get_gateway_url`'s **fallback** branch derives the scheme from
  `connection_info().scheme()`, which consults `X-Forwarded-Proto` / `Forwarded` regardless of peer
  trust. Pre-existing on `main`, unchanged by this PR, and self-targeting (it shapes the QR URL
  returned to the same caller). Not widened into here.
- `rate_limit.rs`'s anonymous-DID key uses `to_string()` on the socket peer without mapped-v4
  canonicalization, so it can still split across two spellings — the #2557 / #2563 / #2564 class.
  It reads the socket peer only, so it is not a forwarded-identity defect.
- The shipped K8s topology exposes the gateway via **NodePort 30080 with no Ingress object**, and
  `externalTrafficPolicy` is unset (defaults to `Cluster`), so kube-proxy SNATs the client address
  to the node regardless of this change. Leaving `trusted_proxy_ips` commented out is therefore
  correct for what is actually shipped; both `configMapKeyRef`s are `optional: true`, so the
  commented key cannot break pod startup.

---

## Final reviewed record

Bookkeeping pass after the independent review. No source changes; this section is the record the
merge gate should rely on instead of re-deriving it.

### Authority invariant, as finally reviewed

> Forwarded client identity is an **untrusted request claim** unless the immediate transport peer is
> explicitly trusted as a proxy. A trusted peer permits *interpretation* of the asserted forwarding
> chain, but traversal proceeds **from the transport-peer side inward and stops at the first
> untrusted hop**.

And the provenance distinction, which is the part that is easy to lose:

> Trusting the transport peer establishes **who may assert** forwarding metadata. It does not make
> the contents true. The proxy must still *construct* that metadata correctly — merely relaying an
> attacker-supplied `X-Forwarded-For` launders the claim through the gate. The appliance nginx
> change completes this contract for the bundled same-host deployment by appending the address
> nginx actually observed.

This is **not** "use the rightmost `X-Forwarded-For` value". Rightmost-wins is a different bug: it
still reads a caller-supplied header without establishing that anyone trustworthy wrote it, and it
breaks with two proxies.

### Mutation evidence (completed — do not rerun)

| | mutation | result |
| --- | --- | --- |
| REPRO | restore the original vulnerable `get_client_ip` | **killed** — 5/6 integration red, all unit green |
| M1 | trust `X-Forwarded-For` from untrusted peers | **killed** |
| M2 | caller-controlled / wrong chain direction | **killed** |
| M3 | ignore forwarding metadata entirely | **killed** |
| M4 | remove strict mapped-v4 normalization | **killed** |
| M5 | consume only the first `X-Forwarded-For` field-line | **killed** |

REPRO is the load-bearing one: every *unit* test stayed green while 5 of 6 *integration* tests went
red. That asymmetry is why the suite enters through the real route rather than the helper.

### Deployment classifications

| deployment | proxy → gateway | trusted by default? | classification |
| --- | --- | --- | --- |
| appliance LAN templates | nginx → `127.0.0.1:8080`, now appends observed client | **yes**, loopback | authority chain **complete** |
| `deploy/config/nginx.conf` (compose) | nginx → `icnd:8080`, appends observed client | **no** — container/network address | **fail-closed**; operator must configure the exact address |
| `deploy/k8s` | none — NodePort 30080, no `Ingress` shipped | n/a | **NOT A DEFECT** |

The application deliberately does **not** broadly trust RFC1918 or container/cluster CIDRs.

### `TRUSTED_PROXY_IPS` semantics

| value | effect |
| --- | --- |
| unset | loopback only (`127.0.0.1`, `::1`) |
| non-empty | the supplied list **replaces** the loopback default — it does not extend it |
| set-and-empty, or every entry unparseable | trusts **nobody** (fail-closed) |

Consequence worth stating plainly: an operator running **both** the same-host appliance nginx **and**
another reverse proxy must list loopback explicitly, or the on-host proxy is silently de-trusted.
Behaviour is unchanged by this PR; it is documented in `client_ip.rs` and in the two nginx configs.

### Same-class audit — settled, do not re-derive

**In scope, fixed:** `api/auth.rs`, `api/listings.rs`, `api/sessions.rs` migrated to one canonical
`client_ip` authority; appliance nginx now asserts `X-Forwarded-For`; duplicate field-line handling.

**Reviewed, not #2568 defects:** `rate_limit.rs:549` uses the raw socket peer and never consumes the
forwarded helper; `/v1/dev/demo/` proxies to a separate loopback-bound demo service that gates on
`Origin`, not client IP; Kubernetes ships NodePort with no ingress.

**Follow-ups filed (not implemented):**

- #2569 — `fix(gateway): authenticate forwarded scheme and host metadata before use`
- #2570 — `fix(gateway): canonicalize socket peer IP before anonymous rate-limit keying`
- #2571 — `refactor(gateway): cache and startup-validate the TRUSTED_PROXY_IPS allowlist`

#2571 came from the Copilot review on this PR and is classified non-blocking: caching changes cost,
not the trust outcome. None of the three are closed by this PR.

### DIFF
--- a/deploy/appliance/lan/nginx-icn-rehearsal-http.conf.in
+++ b/deploy/appliance/lan/nginx-icn-rehearsal-http.conf.in
@@ -35,6 +35,12 @@ server {
         proxy_pass http://127.0.0.1:8080;
         proxy_http_version 1.1;
         proxy_set_header Host $host;
+        # The gateway trusts loopback to assert forwarding metadata, so this proxy MUST
+        # assert it. $proxy_add_x_forwarded_for appends the address nginx actually saw;
+        # the gateway reads the rightmost untrusted hop, discarding anything the caller
+        # prepended. Omitting this line forwards a caller-supplied header verbatim and
+        # launders it through the trust gate (#2567).
+        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
     }
 
     # Static: rehearsal landing + member shell + fixture pack.
--- a/deploy/appliance/lan/nginx-icn-rehearsal-https.conf.in
+++ b/deploy/appliance/lan/nginx-icn-rehearsal-https.conf.in
@@ -44,6 +44,12 @@ server {
         proxy_pass http://127.0.0.1:8080;
         proxy_http_version 1.1;
         proxy_set_header Host $host;
+        # The gateway trusts loopback to assert forwarding metadata, so this proxy MUST
+        # assert it. $proxy_add_x_forwarded_for appends the address nginx actually saw;
+        # the gateway reads the rightmost untrusted hop, discarding anything the caller
+        # prepended. Omitting this line forwards a caller-supplied header verbatim and
+        # launders it through the trust gate (#2567).
+        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
     }
 
     # Static: rehearsal landing + member shell + fixture pack.
--- a/deploy/config/nginx.conf
+++ b/deploy/config/nginx.conf
@@ -10,6 +10,15 @@ server {
     }
 
     # Proxy API requests to icnd
+    #
+    # This asserts X-Forwarded-For correctly ($proxy_add_x_forwarded_for appends the address
+    # nginx actually saw), but the gateway only reads it from a trusted peer. Here the peer is
+    # this container's address on the compose network, not loopback, so it is untrusted by
+    # default: since #2567 X-Forwarded-For is ignored and every client behind this proxy shares
+    # one rate-limit bucket keyed on the container address. That is fail-closed, not fail-open.
+    # To restore per-client attribution, set TRUSTED_PROXY_IPS on the icnd service to this
+    # container's address (exact IPs, no CIDR). Note a non-empty value replaces the loopback
+    # default rather than extending it.
     location /v1 {
         proxy_pass http://icnd:8080;
         proxy_http_version 1.1;
--- a/deploy/k8s/configmap.yaml
+++ b/deploy/k8s/configmap.yaml
@@ -11,10 +11,20 @@ data:
   # Examples: "https://icn.example.com", "http://10.8.30.40:30080"
   gateway_base_url: "http://10.8.30.40:30080"
 
-  # Trusted proxy IPs that can set X-Forwarded-* headers (comma-separated)
-  # Only needed if GATEWAY_BASE_URL is not set and you're behind a reverse proxy
-  # Example: "10.42.0.1,10.43.0.1" (K8s cluster IPs)
-  # If not set, only localhost (127.0.0.1, ::1) is trusted
+  # Trusted proxy IPs that may assert forwarding metadata (comma-separated, exact
+  # addresses, no CIDR). Example: "10.42.0.1,10.43.0.1" (K8s cluster IPs).
+  # If not set, only localhost (127.0.0.1, ::1) is trusted.
+  #
+  # This gates BOTH the QR-login URL derivation and — since #2567 — the client IP used
+  # for rate limiting and audit attribution on unauthenticated endpoints. If you run an
+  # ingress in front of the gateway and leave this unset, that ingress is untrusted:
+  # X-Forwarded-For is ignored and every client behind it shares one rate-limit bucket
+  # keyed on the ingress address. That is fail-closed, not fail-open, but it will limit
+  # real users. Set this to the ingress address before serving traffic through a proxy.
+  #
+  # The proxy must also set X-Forwarded-For itself (nginx:
+  # `proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;`). A trusted proxy that
+  # forwards the client's header verbatim launders a caller-supplied claim through the gate.
   # trusted_proxy_ips: "10.42.0.1"
   icn.toml: |
     data_dir = "/data"
--- a/icn/crates/icn-gateway/src/api/auth.rs
+++ b/icn/crates/icn-gateway/src/api/auth.rs
@@ -11,27 +11,13 @@ use crate::session_authority::SessionAuthority;
 use crate::validation;
 use icn_obs::metrics::gateway;
 
-/// Extract client IP address from request
-/// Returns IP as string, using X-Forwarded-For header if present (proxy support)
+/// Extract client IP address from request.
+///
+/// SECURITY: `X-Forwarded-For` is honoured only when the immediate transport peer is a
+/// trusted proxy; otherwise the socket peer is the client (#2567). See
+/// [`crate::client_ip`] for the trust model.
 fn get_client_ip(req: &HttpRequest) -> String {
-    // Check for X-Forwarded-For header first (reverse proxy support)
-    if let Some(forwarded) = req.headers().get("x-forwarded-for") {
-        if let Ok(forwarded_str) = forwarded.to_str() {
-            // X-Forwarded-For can contain multiple IPs (client, proxy1, proxy2...)
-            // Use the first one (actual client)
-            if let Some(client_ip) = forwarded_str.split(',').next() {
-                return client_ip.trim().to_string();
-            }
-        }
-    }
-
-    // Fall back to connection info
-    if let Some(peer_addr) = req.peer_addr() {
-        return peer_addr.ip().to_string();
-    }
-
-    // Last resort: return placeholder (should never happen)
-    "unknown".to_string()
+    crate::client_ip::client_ip_key(req)
 }
 
 /// POST /auth/challenge - Request authentication challenge
--- a/icn/crates/icn-gateway/src/api/listings.rs
+++ b/icn/crates/icn-gateway/src/api/listings.rs
@@ -27,24 +27,13 @@ use icn_identity::Did;
 // Helper Functions
 // ============================================================================
 
-/// Extract client IP address from request
-/// Returns IP as string, using X-Forwarded-For header if present (proxy support)
+/// Extract client IP address from request.
+///
+/// SECURITY: `X-Forwarded-For` is honoured only when the immediate transport peer is a
+/// trusted proxy; otherwise the socket peer is the client (#2567). See
+/// [`crate::client_ip`] for the trust model.
 fn get_client_ip(req: &HttpRequest) -> String {
-    // Check for X-Forwarded-For header first (reverse proxy support)
-    if let Some(forwarded) = req.headers().get("x-forwarded-for") {
-        if let Ok(forwarded_str) = forwarded.to_str() {
-            // X-Forwarded-For can contain multiple IPs (client, proxy1, proxy2...)
-            // Use the first one (actual client)
-            if let Some(client_ip) = forwarded_str.split(',').next() {
-                return client_ip.trim().to_string();
-            }
-        }
-    }
-
-    // Fall back to peer address
-    req.peer_addr()
-        .map(|addr| addr.ip().to_string())
-        .unwrap_or_else(|| "unknown".to_string())
+    crate::client_ip::client_ip_key(req)
 }
 
 /// Extract caller DID from JWT claims (required auth)
--- a/icn/crates/icn-gateway/src/api/sessions.rs
+++ b/icn/crates/icn-gateway/src/api/sessions.rs
@@ -44,19 +44,14 @@ const WEB_SESSION_SCOPE_CEILING: &[&str] = &[
     "governance:write",
 ];
 
-/// Extract client IP address from request (for rate limiting)
+/// Extract client IP address from request (for rate limiting).
+///
+/// SECURITY: `X-Forwarded-For` is honoured only when the immediate transport peer is a
+/// trusted proxy; otherwise the socket peer is the client (#2567). This is the same
+/// allowlist [`get_gateway_url`] gates `X-Forwarded-*` on, via
+/// [`crate::client_ip::is_trusted_proxy`].
 fn get_client_ip(req: &HttpRequest) -> String {
-    if let Some(forwarded) = req.headers().get("x-forwarded-for") {
-        if let Ok(forwarded_str) = forwarded.to_str() {
-            if let Some(client_ip) = forwarded_str.split(',').next() {
-                return client_ip.trim().to_string();
-            }
-        }
-    }
-    if let Some(peer_addr) = req.peer_addr() {
-        return peer_addr.ip().to_string();
-    }
-    "unknown".to_string()
+    crate::client_ip::client_ip_key(req)
 }
 
 /// Get gateway base URL from environment or request headers
@@ -72,21 +67,12 @@ pub(crate) fn get_gateway_url(req: &HttpRequest) -> String {
         return url;
     }
 
-    // Check if the request came from a trusted proxy
-    let is_trusted_proxy = if let Some(peer_addr) = req.peer_addr() {
-        let peer_ip = peer_addr.ip().to_string();
-
-        // Get trusted proxy list from environment (comma-separated IPs)
-        // Example: TRUSTED_PROXY_IPS="127.0.0.1,::1,10.0.0.1"
-        if let Ok(trusted_ips) = std::env::var("TRUSTED_PROXY_IPS") {
-            trusted_ips.split(',').any(|ip| ip.trim() == peer_ip)
-        } else {
-            // If TRUSTED_PROXY_IPS not set, only trust localhost
-            peer_ip == "127.0.0.1" || peer_ip == "::1"
-        }
-    } else {
-        false
-    };
+    // Check if the request came from a trusted proxy.
+    // One trust model for the whole gateway — see `crate::client_ip` for the semantics of
+    // TRUSTED_PROXY_IPS (exact addresses, comma-separated, loopback by default).
+    let is_trusted_proxy = req
+        .peer_addr()
+        .is_some_and(|peer_addr| crate::client_ip::is_trusted_proxy(peer_addr.ip()));
 
     // Only trust X-Forwarded-* headers if from a trusted proxy
     if is_trusted_proxy {
--- a/icn/crates/icn-gateway/src/client_ip.rs
+++ b/icn/crates/icn-gateway/src/client_ip.rs
@@ -1,0 +1,238 @@
+//! Canonical client-IP derivation for the gateway (#2567).
+//!
+//! # Trust model
+//!
+//! Forwarded client identity is an **untrusted request claim** unless the immediate transport
+//! peer is explicitly trusted to assert forwarding metadata. If the immediate peer is not
+//! trusted, the socket peer IP is the client IP.
+//!
+//! This is the same allowlist `api::sessions::get_gateway_url` uses to gate `X-Forwarded-*`
+//! for QR-login URL derivation; both read it through [`is_trusted_proxy`] so the gateway has
+//! exactly one trusted-proxy model.
+//!
+//! # Configuration
+//!
+//! `TRUSTED_PROXY_IPS` is a comma-separated list of **exact** proxy addresses (no CIDR):
+//!
+//! ```text
+//! TRUSTED_PROXY_IPS="10.42.0.1,::1"
+//! ```
+//!
+//! - unset — loopback only (`127.0.0.1`, `::1`)
+//! - set and empty — trust nobody; every request is keyed on its socket peer
+//! - unparseable entries are ignored (they cannot match a kernel-supplied peer address)
+//!
+//! Addresses are compared as parsed [`IpAddr`]s with IPv4-mapped IPv6 unwrapped, so a peer
+//! arriving as `::ffff:10.42.0.1` on a dual-stack listener matches a `10.42.0.1` entry. Both
+//! sides of the comparison come from the same normalization, so this narrows nothing: the
+//! mapped form denotes the same host. It also keeps the rate-limit key from splitting across
+//! two spellings of one address, the class fixed in #2557 / #2563.
+//!
+//! # Deployment note
+//!
+//! A trusted proxy must *overwrite or append to* `X-Forwarded-For` itself. A proxy that
+//! forwards the client's header verbatim launders the claim through the trust gate: the
+//! gateway cannot tell a proxy-asserted hop from a caller-supplied one.
+
+use std::net::IpAddr;
+
+use actix_web::HttpRequest;
+
+/// Header carrying the forwarding chain. `X-Real-IP` and RFC 7239 `Forwarded` are
+/// deliberately not consulted — neither is set by any proxy config in this repository, and
+/// reading them would widen the attacker-controlled input surface for no deployment benefit.
+const FORWARDED_FOR: &str = "x-forwarded-for";
+
+/// Environment variable holding the trusted-proxy allowlist.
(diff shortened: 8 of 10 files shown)


fahertym:
Valid observation, classified as **FOLLOW-UP (performance/hygiene), non-blocking for this PR** — tracked as #2571.

Why it is not fixed here:

- **No security impact.** `trusted_proxies()` returns the same allowlist either way; caching changes cost, not the trust outcome. The authority boundary this PR exists to fix is unaffected.
- The log-spam half is overstated at production levels: that line is `tracing::debug!`, which is level-filtered, and `tracing` skips field evaluation when the level is disabled. It appears only with debug logging enabled *and* a malformed config.
- Caching is **not** a pure refactor, so it should not be a merge-gate reflex. It stops `TRUSTED_PROXY_IPS` being re-read at runtime (almost certainly fine, but a decision worth stating), and it changes the test seam — this PR's review drove a 43-case matrix through the real `/auth/verify` path by varying `TRUSTED_PROXY_IPS` per case, which a naive `OnceLock` would break.

The cost is real and worth fixing: the parse runs on the rate-limit/audit path for all five migrated routes plus the QR-URL path, and `std::env::var` serializes against other env access.

#2571 also carries the more valuable half hiding inside this report: **validate at startup and log loudly**. Today a CIDR typo like `TRUSTED_PROXY_IPS="10.42.0.0/16"` silently yields an empty allowlist — fail-closed, but the only signal is a `debug!` line while the symptom is "every client shares one rate-limit bucket". That is an operability trap worth converting into a startup warning.

Issue #2469, opened 2026-07-29
fix(governance): authenticated governance replication over the gossip apply path


fahertym:
**P0 security follow-up to the F-P0-2 containment.**

The containment suspends application of remotely-replicated governance state. It does
**not** deliver authenticated governance replication. This issue owns that durable work.
Until it lands, multi-node governance convergence over `governance:proposal` and
`federation:governance:*` is intentionally paused.

## Corrected mechanism

`GossipEntry` (`icn/crates/icn-gossip/src/types.rs`) carries a claimed `author` DID and
**no signature binding that DID to the entry contents**. `GossipActor::store_entry`
(`icn/crates/icn-gossip/src/gossip.rs`) neither recomputes the entry hash (it dedups on
the attacker-supplied `entry.hash`) nor enforces the topic ACL on receipt. The governance
replication ingress therefore received an attacker-chosen DID.

Note for anyone reading older campaign notes: the earlier claim that the bypass was
`sender == None` skipping optional checks is **refuted**. The production call site always
passed `Some(&entry.author)`. The defect was that `entry.author` was never authenticated,
so the arms that compared against it were defeated by impersonation, and four arms
(`DomainCreated`, `DomainUpdated`, `ProposalCreated`, `VoteCast`) compared against nothing
at all. Do not reintroduce the `sender == None` theory.

## Containment boundary already in place

The governance replication ingress (`icn/apps/governance/src/actor.rs`) refuses to apply
any replicated `GovernanceMessage` to governance state, and the replication closure no
longer holds a `GovernanceStateStore` handle at all. Generic gossip transport, storage,
vector-clock merge and anti-entropy are unchanged. Local governance operations are
unaffected — they persist through the command path before publishing.

## Required cryptographic and institutional invariants

A durable design must produce a verifiable binding among **all** of:

- message contents
- the institution / governance domain the message affects
- the signing DID
- the network sender or replicated origin
- the authority scope that DID holds over that domain
- a replay identity

Specifically:

1. **Content binding.** A signature over canonical message bytes, verified before any
   state application. Entry hash must be recomputed and checked against content.
2. **Authority binding.** A valid signature is not authority. The verifier must establish
   that the signer held a *specific* governance authority over the *specific* affected
   domain at the relevant time. Note `GovernanceProofV2::verify_receipt` /
   `validate_secure_v2_proof_for_proposal` verify attestation signatures but do **not**
   check signer authority over the domain — an attacker can satisfy them with
[...]

Issue #2441, opened 2026-07-19
fix(federation): unauthenticated institutional-state overwrite via gossip apply paths


fahertym:
**P1 security.** Institutional state is applied from gossip without authority verification, in the live composition:
- Coop records (charter_id, bylaws, tiers, founding signatures): `handle_coop_update` (`icn-core/src/supervisor/init_notifications.rs:720`) decodes and `save_cooperative`s **unconditionally** (:735) — arrival-order wins, no authorization, wired via `init_coop.rs:87`.
- Entity registry (incl. memberships): wall-clock LWW (`icn-entity/src/actor.rs:312-322`) — any peer with a larger `updated_at` wins.
- Gen-1 federation agreements: `handle_signature` checks party-listing only, never verifies signature bytes (`manager.rs:656` pushes `vec![]` with a "would verify properly" comment), and **auto-activates on full signatures from unauthenticated messages** (`agreement/gossip.rs:257-260`; status overwrite :307-309). Handler registered in `init_federation.rs:196-198` — live.
Contrast: **outbound** federation ops are receipt-bound (decision_hash lineage, #2094/#2095); inbound is ambient. ConflictResolver and ledger ForkDetector/ForkResolver exist but have zero runtime callers.

**Fix direction:** typed per-state-class `ReplicationPolicy` enforced at the topic-registration seam; first slice = quarantine (hold + steward release valve) instead of apply for unauthenticated institutional writes, and reject empty-signature agreement messages. **Eyes open:** Gen-1 carries no origin-domain receipts, so quarantine effectively pauses inbound institutional-state sync until receipt-carrying sync exists; freeze new Gen-1 agreement features meanwhile.

Found by the 2026-07-19 architecture verification campaign (federation lane, spot-verified). Refs #2094 #2095.

Issue #2579, opened 2026-08-13
fix(ci): cargo-audit fails on main and skips both cargo-deny gates, so advisory AND license checks never run


fahertym:
The `Security Audit` job in the CI workflow has been failing on `main` itself since at least `99c62311` (2026-08-11), so its signal is dead: every PR inherits a red advisory check that carries no information about the PR.

Noticed while merging #2575 / #2577 / #2578 — each had to classify the same red check as pre-existing before merging. That classification is currently manual and repeated per PR.

## Evidence

The failing step is `Run cargo-deny advisories check` (note `Run cargo-audit` in the same job **succeeds**). On the latest `main` CI run at `57879743`:

```
error[unsound]: Potential use-after-free due to lack of panic safety in `LruCache::pop()`
  ID: RUSTSEC-2026-0253
  Advisory: https://rustsec.org/advisories/RUSTSEC-2026-0253
```

`RUSTSEC-2026-0098` and `RUSTSEC-2026-0099` (rustls-webpki, transitive via reqwest) are already in the ignore list and are not the cause. **RUSTSEC-2026-0253 is not ignored and not tracked by any open issue** — #2101 covers the separate PQClean/pqcrypto migration, not this.

Job conclusion on recent `main` CI runs:

| main SHA | Security Audit |
|---|---|
| `57879743` | failure |
| `77878d71` | failure |
| `b498f9c9` | failure |
| `d34413dd` | success |
| `61ca91d0` | success |

So it turned red between `d34413dd` and `b498f9c9`, consistent with a newly published advisory rather than a repo change.

## Why it is worth a tranche

A permanently-red non-required check trains reviewers to ignore it, which is exactly how a *real* advisory would be missed. The job is doing its job — the finding is genuine — but nothing is acting on it.

## Shape, not prescribed

Decide, with evidence, between:

- upgrading the dependency that pulls in the affected `lru` version (identify it first — `cargo tree -i lru`);
- adding a **dated, justified** ignore entry alongside the existing rustls-webpki ones, with the reasoning recorded, if the unsound path is not reachable from ICN code;
- or splitting advisories into an informational job so a red advisory is visible without being indistinguishable from a broken gate.

Constraints: do not upgrade the pinned toolchain; keep `Cargo.lock` changes in a Rust-only commit; do not silently widen the ignore list without recording why each entry is safe.

Refs #2101

fahertym:
[...]

PR #2582, opened 2026-08-13
feat(governance): add the authenticated replication envelope primitive (#2469)

fahertym:
**Slice 1 of authenticated governance replication. The primitive only — deliberately unwired.**

Nothing here reads or writes governance state, touches gossip, changes migration behavior, or lifts the #2470 containment. `SignedGovernanceOp` is a library type with unit tests and zero callers.

Design artifact: `docs/design/authenticated-governance-replication.md` (added here).

## What this adds

`SignedGovernanceOp` binds a governance operation to the DID that authored it, the domain it affects, and the membership snapshot under which that DID held authority — plus canonical encoding, signing, verification, a derived `op_id`, and a deterministic `static_membership_hash`.

## Why the signature is in the content, not the transport

`SignedEnvelope` authenticates a **hop**, not an author. Verified on current main: `handle_response` (`icn-gossip/src/handlers/push.rs:109`) stores a received entry verbatim, and `handle_request_missing` (`pull.rs:461`) re-serves `entry.clone()` onward under the **relay's own** envelope signature. `entry.author` therefore crosses every hop unauthenticated.

Requiring `envelope.from == entry.author` would accept only single-hop delivery, breaking relay and anti-entropy — destroying the convergence #2469 exists to restore. Authenticating the author requires a signature *inside* the replicated bytes.

## Decisions frozen into the wire format

| Decision | Rationale |
|---|---|
| `op_bytes` signed and transmitted **verbatim** | Serialization determinism is never relied upon — stronger than canonicalizing `GovernanceMessage`, and avoids a hand-rolled encoder over ~20 nested variants where a newly added field would be silently unsigned. |
| Envelope fields canonically encoded | Fixed domain separator, length-prefixed throughout, stable wire discriminants (pinned by test). |
| `op_id` **derived**, not carried | Cannot disagree with content. Order-independent, so it never discards a legitimately unseen op. |
| `seq` is a **comparator, never a gate** | A watermark (`reject seq <= last_seen`) would permanently drop op 10 when op 11 arrived first — which gossip permits and anti-entropy makes routine. |
| Authority binds **membership only** | Quorum/threshold/emergency settings govern how state is *evaluated*, not who may *author*. Binding the whole `GovernanceConfig` would invalidate in-flight votes on an unrelated quorum edit. |
| `sign()` refuses an author/key mismatch | A public constructor must not emit an object that is unverifiable by construction. |

## The v1 authority boundary (corrected in `3b7bab5a` — please read this part)

An initial draft of this work claimed `VoteCast` was restorable on StaticList membership alone. **That was wrong**, and the correction is the most important thing in this PR.

In the production composition, casting a vote *also* requires that the voter is not suspended: `cast_vote` consults a `SuspensionChecker` wired to `CoopManager::is_member_suspended` (`icn-gateway/src/server.rs:1904`), and `close_proposal` revalidates commons Member standing for every voter (`actor.rs:2149-2189`).

**A suspended member remains in the domain's `StaticList`** — precisely the shape `icn-gateway/tests/e2e_close_time_revalidation.rs` constructs. So a suspended-but-listed DID would have passed every proposed ingress check and had a vote applied that the HTTP path refuses with 403: a gossip bypass of a live production gate.

Close-time revalidation does not rescue this. It filters the effective tally, but the vote is stored and observable before close, revalidation is handler-driven and fails open when unwired, and stored state still diverges across nodes.

The predicate **cannot be evaluated on a receiving node**: it lives in commons/coop state, is invisible to `GovernanceActor` (the checkers are HTTP-layer closures), and its own replication is the unauthenticated class tracked by #2441.

> **v1 boundary = `StaticList` membership AND no external authority predicate applies on the receiver.**
> Production compositions **stay contained**. Authenticated `VoteCast` application is not restorable in production in v1.

A `standing_hash` field was considered and **rejected**: the receiver would compare it against unauthenticated commons state, manufacturing false determinism over untrusted data — worse than no field, because it would look like the hole was closed.

**This did not change the field set.** It narrows what slice 7 may apply, not what slice 1 must sign.

## Other compatibility limitation

`TrustThreshold` membership resolves against each node's **local** trust graph, so two honest nodes can reach different verdicts — permanent divergence. It stays fail-closed. `GovernanceConfig::cooperative_default()` uses `trust_threshold(0.3)`, so the default cooperative profile is not eligible for v1 either.

## Risk

Minimal. No production code path reaches this module — zero callers. The only non-additive change is one `pub mod` line in `icn-governance/src/lib.rs`.

## Test plan

34 unit tests: sign/verify round-trip; author/key mismatch refused at construction; adversarial wrong-author envelope still rejected by the verifier (before and after a decode round trip); tamper detection on every signed field; truncated signature; `op_id` determinism and per-field sensitivity; `membership_hash` set-order/duplicate invariance, domain separation, boundary-shift resistance; frame round-trip; bad magic; unknown version; truncation at every prefix length; trailing bytes; over-declared length prefix; unknown discriminants; payload/kind disagreement; legacy bare-JSON payload not mistaken for a frame.

```
cargo fmt --all --check                                        → 0
cargo clippy -p icn-governance --all-targets -- -D warnings    → 0
cargo test -p icn-governance      → 733 + 42 + 4 + 7 passed, 0 failed
doc_control_check.py                                           → 0
```

## Rebased onto `3d401dce` — slice 2 has landed

This branch is rebased onto current main, which now contains **#2583 (slice 2)**: gossip re-derives `entry.hash` from the payload before an unseen entry may claim a content-addressed slot. The two PRs were file-disjoint (3 files here, 8 there, zero overlap), so the rebase was a conflict-free replay — all three blobs are byte-identical to the pre-rebase commits.

`5e0ec94f` re-derives the design doc against `3d401dce`: §1.2/§1.3 no longer claim the hash is never recomputed, T3 moves Open → Closed, §5.6 step 6 becomes an upstream guarantee, and the slice table marks slice 2 done. **No architectural premise moved** — #2583 authenticates payload↔digest only, never authorship, so the containment and the authority boundary below are untouched.

It also corrects a mis-citation in §7.0: revision 2 cited the vote suspension gate at `handlers.rs:875`, which is the **proposer** gate on proposal submission. `handlers.rs` is untouched by #2583, so this was wrong when written rather than drift. The substance is unchanged and **stronger** — `cast_vote` carries *two* external predicates, not one: a `MemberStandingChecker` (`handlers.rs:1850`) and a `SuspensionChecker` (`handlers.rs:1864`).

## Not in this PR

Slices 3–7: signed emission, quarantine, the `op_id` applied-set and comparator, the lifecycle guard, and finally verification + apply. (Slice 2, entry-hash recomputation, landed separately as #2583.) **Only slice 7 lifts containment**, must prove positive convergence through a real relay hop, and must fail closed wherever a standing predicate is wired.

Refs #2469 #2441

### DIFF
--- a/docs/design/authenticated-governance-replication.md
+++ b/docs/design/authenticated-governance-replication.md
@@ -1,17 +1,26 @@
 # Authenticated governance replication (#2469)
 
 **Status:** design + slice 1 (unwired primitive).
-**Derived against:** `origin/main` = `c378232132139235e2a8cc305792eaee2571267a`.
+**Derived against:** `origin/main` = `3d401dce0c65d70eea117335b48719baa2231e2e`.
 **Owns:** the durable replacement for the #2470 containment.
-**Refs:** #2469, #2441, #2470 (`bc291305`), #2471, #2480, #2510, #2520, #2535, #2544.
+**Refs:** #2469, #2441, #2470 (`bc291305`), #2471, #2480, #2510, #2520, #2535, #2544, #2583.
 
-Everything in §1 was re-derived from source at `c3782321`. Where an older campaign note
-disagrees, the source wins and the disagreement is called out.
+Everything in §1 was re-derived from source — originally at `c3782321`, re-verified at
+`3d401dce`. Where an older campaign note disagrees, the source wins and the disagreement is
+called out.
 
 **Revision 2** corrects three mechanisms from revision 1 that were unsafe under
 eventually-consistent delivery: the sequence gate, first-writer-wins collision
 arbitration, and whole-config authority binding. Revision 1's `prev` hash-chain field is
 also removed. See §10, §5.2, §5.3.
+
+**Revision 3** re-derives §1 against `3d401dce`, where **slice 2 has landed** as #2583:
+gossip now re-derives `entry.hash` from the payload before an unseen entry may claim a
+content-addressed slot. T3 is closed, and §5.6 step 6 became an upstream guarantee rather
+than work this design still has to schedule. **No architectural premise changed** — #2583
+authenticates payload↔digest only, and leaves authorship, authority and the #2470
+containment exactly as they were. Revision 3 also corrects a §7.0 mis-citation: the vote
+suspension gate was cited at the *proposer* gate's line.
 
 ---
 
@@ -24,8 +33,8 @@ also removed. See §10, §5.2, §5.3.
 | 1 | Operator issues a `GovernanceCommand` (RPC/HTTP) | `icn-rpc/src/handler/governance.rs`, `apps/governance/src/http/handlers.rs` |
 | 2 | `GovernanceActor` **persists to `GovernanceStateStore` first** | `apps/governance/src/actor.rs` command arms |
 | 3 | Actor builds a `GovernanceMessage`, serializes with `to_bytes()` = **`serde_json::to_vec`** | `icn-governance/src/message.rs:417` |
-| 4 | `gossip.publish(topic, data, author = own_did)`; ACL checked **for the local DID only**; `hash = Self::hash_data(&data)` | `icn-gossip/src/gossip.rs:911`, `:919` |
-| 5 | `store_entry(entry)` → fires notification callbacks (including governance's own loopback copy) | `icn-gossip/src/gossip.rs:968` |
+| 4 | `gossip.publish(topic, data, author = own_did)`; ACL checked **for the local DID only**; `hash = Self::hash_data(&data)` | `icn-gossip/src/gossip.rs:858`, `:925` |
+| 5 | `store_entry(entry)` → fires notification callbacks (including governance's own loopback copy) | `icn-gossip/src/gossip.rs:974` |
 | 6 | Outbound send wraps the `GossipMessage` in a **`SignedEnvelope`** — `PayloadType::Gossip`, Ed25519 by the node keypair, **durable monotonic sequence**, fail-closed if the sequence cannot be persisted (#2510) | `icn-core/src/supervisor/init_send_callback.rs:163` |
 | 7 | Sent as `MessagePayload::Signed` | `icn-net/src/protocol.rs:628` |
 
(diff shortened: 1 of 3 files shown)


fahertym:
## CI triage for `24dad515` — both failures classified as base-wide, no code change made

Two required checks are red. Neither is caused by this PR. Evidence below; **no fix applied, deliberately**.

### `Security Audit` — pre-existing on `main`

`cargo deny check advisories` reports `RUSTSEC-2026-0253` (`lru 0.16.3`, fixed in `>= 0.18.2`).

The same advisory fires on `main` at this PR's own base:

| | run | advisory | Cargo.lock line |
|---|---|---|---|
| `main` @ `c3782321` | [31711191860](https://github.com/InterCooperative-Network/icn/actions/runs/31711191860) job 94484395256 | `RUSTSEC-2026-0253` | `356 │ lru 0.16.3` |
| PR @ `24dad515` | [31718973924](https://github.com/InterCooperative-Network/icn/actions/runs/31718973924) job 94511024813 | `RUSTSEC-2026-0253` | `356 │ lru 0.16.3` |

Identical advisory, identical package version, identical lockfile line. `main` additionally carries 8 more advisories (`RUSTSEC-2023-0089`, `2025-0141`, `2026-0097/0098/0099/0161/0162/0163/0166/0173`), so this PR strictly *narrows* nothing and *adds* nothing.

**This PR changes no dependency files at all** — `git diff --name-only origin/main..HEAD` returns exactly three paths, none of them `Cargo.toml`, `Cargo.lock`, or `deny.toml`. It is mechanically incapable of introducing an advisory. Not fixing it here; that is #2579's scope (`Security Audit` is currently a dead signal on `main`).

### `Test` — pre-existing `icn-net` flake, structurally impossible to be caused by this PR

Failing test: `a_fused_accept_cancelled_after_one_poll_destroys_the_handshake`
(`crates/icn-net/tests/accept_handshake_cancellation.rs:180`).

**Structural proof it cannot be this PR:**
- `cargo tree -p icn-net` contains **zero** occurrences of `icn-governance`. `icn-net` does not depend, directly or transitively, on the only crate this PR touches.
- `git diff origin/main..HEAD -- icn/crates/icn-net/` is **empty**. The failing test and its entire crate are byte-identical to `origin/main`.

**Timing evidence — it fails by being too *fast*, not by regressing.** The assertion's premise is that a one-poll budget cannot span a QUIC/TLS handshake; when the handshake completes quickly the premise collapses and the probe self-reports.

| | duration (7-test file) | result |
|---|---|---|
| CI failing run | **2.14s** | FAILED |
| local, 6× single test | **5.01s** each | 6/6 pass |
| local, 5× full file | **5.01s** each | 5/5 pass |

11/11 local passes, every one at 5.01s. I could **not** reproduce the failure locally — an idle VM consistently lands in the slow/passing regime; the fast regime appears under runner contention. Stating that plainly rather than claiming a repro I did not get.

Corroboration: the `Test` job is independently flaky on `main` — run [31651381839](https://github.com/InterCooperative-Network/icn/actions/runs/31651381839) (`main` @ `57879743`) failed `Test` on a *different* test, `blob_store::tests::blob_store_persistence_across_restart`.

This flake is previously documented on `main` at base `78a01e48` (2 fail / 1 pass with a branch applied, 1 fail / 3 pass with the same changes stashed). It is not filed as an issue yet and probably should be.

### Scope audit

`git diff --stat origin/main..HEAD` → **3 files, 1808 insertions, 0 deletions**:
- `docs/design/authenticated-governance-replication.md`
- `icn/crates/icn-governance/src/lib.rs` (one `pub mod` line)
- `icn/crates/icn-governance/src/replication.rs`

Confirmed: no dependency files changed · **zero** production callers of `SignedGovernanceOp` (grep across the workspace excluding the module itself) · zero changes under `icn/apps/governance/`, `icn/crates/icn-gossip/`, `icn/crates/icn-core/`, `icn/crates/icn-net/` · #2470 containment intact (`observe_replicated_governance_message` still at `actor.rs:3841`, both `F-P0-2 containment` markers present) · no gossip, state-application, or migration behavior changed · `TrustThreshold` unsupported · production `VoteCast` remains contained wherever external standing/suspension authority exists.

### Local gates on `24dad515`

`cargo fmt --all --check` → 0 · `cargo clippy -p icn-governance --all-targets -- -D warnings` → 0 · `cargo test -p icn-governance` → 733 + 42 + 4 + 7 passed, 0 failed · `doc_control_check.py` → 0 · `drift-check.sh` → PASS.

CI agrees on the parts it can see: `Clippy`, `Format Check`, `Build Release`, `Change Scope`, `Meaning Firewall Check`, `Kernel Forbidden Dependencies`, `Documentation Control Plane` all green.

### Status

Staying **draft**. Both red checks are base-wide and outside this PR's scope; fixing either here would violate the scope boundary and add dependency churn to a PR that deliberately contains none. Ready for review on content.

PR #2652, opened 2026-08-23
ops(icn-dev): make the disk guard reproducible and boot-persistent

fahertym:
Host-ops tooling for the `icn-dev` development VM. No product code, no CI changes.

## Incident

Every agent worktree under `~/icn-dev/worktrees/<repo>/<name>` builds into its own Cargo
`target/`. Each is 3–26 GB, and a full workspace build across feature sets can exceed 100 GB.
Nothing ever reclaimed them. On **2026-08-23** the root filesystem reached **96% (20 GB free)**
carrying **372 GB of build output across 19 worktrees** — one worktree alone held 119 GB.

Builds had already started failing with `Blocking waiting for file lock on artifact directory`.

## Immediate recovery (already done; not part of this diff)

**196 GB reclaimed, 96% → 56%.** Nine `target/` directories removed, each individually gated on
merged-PR state, clean tree, zero process cwds, zero open fds, valid `CACHEDIR.TAG`, and no
non-Cargo content. Afterwards every touched worktree was verified to retain its source, HEAD and
branch. **0 branches deleted, 0 worktrees removed, 0 git mutations**; 39 worktrees still
registered.

## What this PR adds

The guard was host-local, so a rebuild of `icn-dev` would silently lose the protection.

| File | Purpose |
|---|---|
| `ops/scripts/icn-disk-guard` | The guard; all policy in one CONFIG block |
| `ops/scripts/icn-disk-guard.service` | System-level oneshot, runs as the dev user |
| `ops/scripts/icn-disk-guard.timer` | Boot-persistent, every 4 h |
| `ops/scripts/install-icn-disk-guard.sh` | Idempotent installer |
| `ops/scripts/README-icn-disk-guard.md` | Policy, rationale, and the gap it does not close |

## Policy: three layers

Filesystem percentage is a **lagging** signal here. A single worktree reached 119 GB, so this
box can go from comfortable to critical inside one agent session. An earlier iteration only
cleaned above 80% and would have sat silent through exactly the run that caused the incident.

**1. Routine** — runs at any disk level. Reclaims output that is `MERGED-INACTIVE`, unpinned,
clean, **≥ 7 days idle and ≥ 5 GB**. This is what actually bounds growth.

The 5 GB floor is chosen from the observed distribution rather than picked: across 19 measured
targets the median is 7.2 GB, so a 10 GB floor sits *above* the median and covers 6/19 targets
(308 GB), where 5 GB covers 13/19 (355 GB). The incident came as much from many mid-sized
targets as from the two outliers, and a routine clean's safety rests on the `MERGED-INACTIVE`
classification, not on the size floor — the floor only avoids churning on trivia.

**2. Pressure** — above **78%** the floors drop to 2 days / 1 GB. It relaxes *staleness and size
only*. Identity, liveness and git-safety checks are never relaxed; five self-tests pin that.
Critical at **90%**.

**3. Budgets** — the resource itself, independent of both: total footprint (**warn 150 GB / crit
200 GB**) and largest single target (**warn 50 GB / crit 100 GB**). These raise the exit code, so
a scheduled run is loud *before* the filesystem is endangered.

> Live proof this layer matters: the guard currently **exits 2 (CRITICAL) at 56% disk**, because
> one target is 113 GB. The previous design exited 0 and printed nothing.

All ten thresholds are declared exactly once and are environment-overridable so tests can scale
them; **0 hardcoded threshold comparisons remain in the logic**, and two self-tests prove it by
overriding a variable and asserting the verdict changes.

## Process-pinned: reported, never acted on

A merged worktree still held by a live process is classified `MERGED-BUT-PROCESS-PINNED` and
reported with PIDs, command and age. The guard **never terminates a process** to make a target
eligible.

Pinned-ness deliberately does **not** raise the exit code on its own: long-lived agent shells in
merged worktrees are the steady state on this host, so escalating there would warn every 4 hours
forever and train an operator to ignore it. The case that matters is already covered — a pinned
target still counts toward both budgets, so a large one trips them anyway, and this block then
explains why it cannot be reclaimed.

## Fail-safe

If `gh` is missing, unauthenticated or rate-limited, PR state is unknown → `STALE-UNKNOWN` →
nothing is removed. Detached worktrees name no branch and are likewise `STALE-UNKNOWN`.

Verified the sandbox does not silently cause this: `gh` was run inside the exact
`ProtectSystem=strict` unit sandbox and returned `MERGED`, and a real timer run produced a
genuine mix of `OPEN-PR` / `MERGED` / `STALE-UNKNOWN` rather than uniform unknowns.

## Why a system timer, not user linger

A user timer only runs while a login session exists, and this host has `Linger=no` — it would
stop silently after an unattended reboot. Enabling lingering was the alternative, but auditing
the blast radius first showed it would also start `dotfiles-sync.service`, which runs
`git pull --ff-only`, unattended at boot. One system unit running as the dev user gives
boot-persistent housekeeping with a far smaller blast radius.

**Cadence:** `OnBootSec=15min` + `OnUnitActiveSec=4h` + `RandomizedDelaySec=20min`. Relative
rather than `OnCalendar`, so runs never synchronise to a wall-clock instant and never stack up
after downtime.

## Why not a shared `CARGO_TARGET_DIR`

Investigated and rejected on evidence. `sccache` already provides cross-worktree compile reuse —
**79.5% hit rate**, `SCCACHE_DIR=/mnt/build/sccache`. What per-worktree `target/` duplicates is
link output and incremental state, which sccache does not dedupe.

A shared target directory would be actively harmful: Cargo takes an **exclusive lock per target
directory**, so ~19 concurrent agent worktrees would serialise every build. This was observed
live — two invocations from a *single* session already hit
`Blocking waiting for file lock on artifact directory` with separate targets. Differing feature
sets (`post-quantum` vs default) would thrash it further.

## Evidence

**Self-tests — 40/40.** Including a positive control, so the suite cannot pass by refusing
everything; five tests that pressure does not relax identity/liveness/git checks; budget
escalation ordering; and `STALE-UNKNOWN` / empty-class fail-safes.

**Synthetic deletion, end to end.** A 6 GB, 30-day-old target was created in a merged, clean,
unpinned worktree while `/` sat at 57% — well below the 78% pressure threshold. The guard
classified it `MERGED-INACTIVE → ELIGIBLE:routine` and `--apply` removed it, leaving all 18
source entries, HEAD, branch, `.git` and the worktree registration intact:

```
removed size=6.1G path=.../task-n1-authority-log/icn/target
  class=MERGED-INACTIVE layer=routine reason=stale-merged-build-output
```

**Quiet-mode.** Verified silent when healthy (0 bytes, exit 0) and fully loud despite `--quiet`
when not (exit 2). An earlier iteration printed eagerly and a `--quiet` timer run stayed silent
at CRITICAL — correct exit code, empty journal. Output is now buffered and flushed by an EXIT
trap so `--quiet` can be revoked retroactively once the budgets are known.

**Checks.** ShellCheck clean, `bash -n` clean, installer idempotent (re-run over a live install
produces an identical unit), `git diff --check` clean, drift/doc-control/compliance/overclaim/
meaning-firewall/skill-registry all pass.

## Known limitation — stated plainly

**This does not bound build output belonging to an active, open-PR, or process-pinned worktree.**
Those are protected by design, at any size.

That limit is load-bearing today: a *merged* lane (`task-2640-respelling-replay`, PR #2644)
currently holds **114 GB** the guard refuses to touch, because shells with a cwd inside it are
still running — some of them days old. The guard reports it, trips per-target CRITICAL on it, and
stops there.

Bounding that requires retiring stale agent sessions, which is agent-lifecycle work and
deliberately outside this tool's remit. Tracked separately.

## Known gap — reprovisioning

This repository has no dev-host provisioning subsystem: `deploy/appliance/systemd/` is the
shipped *product* appliance, not the dev VM, and there is no `ansible/`, `provision/` or `hosts/`
tree. These files therefore sit in `ops/scripts/`, following `setup-skill-symlinks.sh`, which is
likewise tracked-in-repo tooling that configures machine-local developer-host state.

**Consequence: reprovisioning `icn-dev` will not install this automatically** — the installer must
be run by hand. Documented rather than papered over by inventing a subsystem for one guard.

## Non-goals

No product code. No CI changes. No Docker/journal/package-cache pruning. No worktree or branch
deletion. No shared `CARGO_TARGET_DIR`.

### DIFF
--- a/ops/scripts/README-icn-disk-guard.md
+++ b/ops/scripts/README-icn-disk-guard.md
@@ -1,0 +1,103 @@
+# icn-dev disk guard
+
+Bounds Cargo build-artifact growth on the `icn-dev` development VM.
+
+## The incident this exists to prevent
+
+Every agent worktree under `~/icn-dev/worktrees/<repo>/<name>` builds into its own Cargo
+`target/`. Each is 3–26 GB, and a full workspace build across feature sets can exceed 100 GB.
+Nothing ever reclaimed them. On **2026-08-23** the root filesystem reached **96% (20 GB free)**
+with **372 GB of build output across 19 worktrees** — one worktree alone held 119 GB.
+
+## Files
+
+| File | Purpose |
+|---|---|
+| `icn-disk-guard` | The guard. All policy lives in one CONFIG block at the top. |
+| `icn-disk-guard.service` | System-level oneshot, runs as the dev user. |
+| `icn-disk-guard.timer` | Boot-persistent, every 4 h with randomised delay. |
+| `install-icn-disk-guard.sh` | Idempotent installer; refuses to schedule a guard failing its own tests. |
+
+```bash
+ops/scripts/install-icn-disk-guard.sh            # install or update
+ops/scripts/install-icn-disk-guard.sh --check    # report state, change nothing
+icn-disk-guard                                    # audit; dry-run, never deletes
+icn-disk-guard --self-test                        # 40 policy assertions
+```
+
+## Policy
+
+Three layers, because filesystem percentage is a *lagging* signal — one worktree reached
+119 GB on its own, so this disk can go from comfortable to critical inside a single session.
+
+1. **Routine housekeeping** — runs whatever the filesystem looks like. Reclaims Cargo output
+   that is `MERGED-INACTIVE`, unpinned, clean, **≥ 7 days idle and ≥ 5 GB**. This is what
+   actually bounds growth.
+2. **Pressure housekeeping** — above **78%** the age/size floors drop to 2 days / 1 GB. It
+   never relaxes the identity, liveness or git-safety checks; only how stale and large the
+   output must be. Critical at **90%**.
+3. **Budgets** — reports on the resource itself: total footprint (warn **150 GB**, crit
+   **200 GB**) and largest single target (warn **50 GB**, crit **100 GB**). These raise the
+   exit code, so a scheduled run is loud *before* the filesystem is endangered.
+
+The 5 GB routine floor is chosen from the observed distribution: across 19 measured targets
+the median is 7.2 GB. A 10 GB floor sits above the median and would cover 6/19 targets
+(308 GB); 5 GB covers 13/19 (355 GB). The incident was caused as much by many mid-sized
+targets as by the two >100 GB outliers, and a routine clean's safety rests on the
+`MERGED-INACTIVE` classification, not on the size floor.
+
+## What it will not do
+
+Removes only directories named `target/` that carry a Cargo `CACHEDIR.TAG`, and only from
+worktrees positively classified `MERGED-INACTIVE`. Never deletes source, git metadata,
+branches or worktrees; never prunes Docker or package caches; never mutates git.
+
+**Fails safe in every unknown.** If `gh` is missing, unauthenticated or rate-limited, PR state
+is unknown, the worktree classifies `STALE-UNKNOWN`, and nothing is removed.
+
+**Never terminates a process.** A merged worktree still pinned by a live process is reported
+as `MERGED-BUT-PROCESS-PINNED`, with PIDs, command and age. That is agent-lifecycle debt for
+an operator, not something a disk guard should resolve by killing things.
+
+## What this does NOT bound
+
+State plainly, because the guard's name invites the wrong assumption: **it cannot bound build
+output belonging to an active, open-PR, or process-pinned worktree.** Those are protected by
+design and at any size.
+
+That limit is load-bearing right now, not hypothetical. At the time of writing a *merged* lane
+(`task-2640-respelling-replay`, PR #2644) holds **114 GB** that the guard refuses to touch
+because shells with a cwd inside it are still running — some days old. The guard reports it,
+raises the per-target CRITICAL budget on it, and stops there.
+
+So the guard bounds *reclaimable* output. Bounding the rest requires retiring stale agent
+sessions, which is an agent-lifecycle concern and deliberately not in this tool's remit.
+
+## Why a system unit rather than a user unit
+
+A user timer only runs while a login session exists, and this host has `Linger=no` — so it
+would silently stop after an unattended reboot. Enabling lingering was the alternative, but on
+this host that would also start `dotfiles-sync.service` (a `git pull`) unattended at boot,
+which is outside this guard's remit. One system unit running as the dev user gives
+boot-persistent housekeeping with the smallest blast radius.
+
+## Known gap
+
+This repository has no dev-host provisioning subsystem. `deploy/appliance/systemd/` is the
+shipped *product* appliance, not the development VM, and there is no `ansible/`, `provision/`
+or `hosts/` tree. These files therefore sit in `ops/scripts/`, the narrowest existing location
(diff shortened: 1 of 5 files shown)


fahertym:
Re-evaluated against the runtime truth that now exists (#2671, `cd943e02`). **Premise confirmed; the core safety invariant must not be relaxed.**

**Premise confirmed.** Retiring only lanes with positive merge evidence reclaimed **121 GB** today (181G → 60G, 54 → 33 worktrees). The disk pressure this PR targets is real and is mostly Cargo output behind merged lanes nobody retired.

**What improved.** The session registry is no longer blind. Before #2671 it reported zero sessions while four Claude activations were live, so *every* lane read as REGISTRY-UNAVAILABLE and the guard's protected-by-default stance was doing all the work. A registered session is now a real signal.

**What has NOT changed, and must not.** `missing registry row != permission to delete` still holds, for two reasons observed directly today:

1. **Orphans never register.** The 16 abandoned load-generator processes (`ppid=1`) and a `sleep 999999` pinning `task-agent-runtime` are not sessions and never will be. No registry row would ever have existed for them.
2. **A live owner can hold a lane with no row for it.** `task-2645-replay-cleanup-holds` is pinned by a shell whose parent Claude activation is alive, and `task-readiness-audit` carries a row whose heartbeat is five hours stale *because that lane still runs the pre-fix hook and cannot record heartbeats*. A stale or absent row is as likely to mean "the lane could not report" as "nothing is there."

So the ordering stands: **live process pin > registry row > everything else**, uncertainty fails safe, and cleanup never kills a process to reclaim a worktree. If this lands, consume `icn-agent-session classify` for registered sessions rather than re-deriving liveness, and treat REGISTRY-UNAVAILABLE as protected — which is what it already does.

Not implemented or merged here; next-action note only.

fahertym:
## Disposition: **KEEP_DRAFT** — reconciled against `main` at `3c1e4f863` (2026-09-06)

**Not superseded, and the old-numbers concern does not apply here.** I checked specifically for the failure mode where a dev-host PR bakes in historical facts:

- Paths are parameterized through `ICN_DEV_ROOT` / `$HOME/icn-dev` — no hardcoded host addresses.
- Thresholds are named configuration variables, not literals buried in logic.
- The numeric values that do appear in the diff are **test fixtures**, not asserted facts about any current host.

So this is not carrying a stale 2026-08 snapshot of the VM forward as truth.

**Nor is the solution shape superseded.** Newer session infrastructure — `.claude/hooks/session-lifecycle.sh` and the `ops/mcp` session tools — covers *session* lifecycle: registration, heartbeat, release. This pair covers *worktree and disk* debt. Non-overlapping; neither makes the other redundant.

**What actually blocks these is not code — it is a placement decision.** Both are dev-VM operations tooling living in the product repository, and both have **zero reviews**. The open question a maintainer must answer is whether `ops/` in this repo is the right home for icn-dev host hygiene, or whether it belongs in the operations repo.

**#2652 and #2653 should be decided together** — they are the same shape, the same scope question, and the same author, and splitting the decision would leave one orphaned.

They are read-only, valid, uncontested, and cost nothing to leave as drafts. Conflict risk **LOW** (zero competing commits since merge-base `abdfab79`). Keeping them as drafts pending that placement call is the honest state — not a rebase problem, and not a close.

PR #2653, opened 2026-08-23
ops(icn-dev): report agent-worktree lifecycle debt (read-only)

fahertym:
Read-only lifecycle auditor for `icn-dev` agent worktrees. Host-ops tooling; no product code.
Companion to #2652 — **deliberately not folded into it**, and the two tools stay decoupled.

## Problem

`icn-disk-guard` (#2652) refuses to reclaim a Cargo target while any process holds the
worktree. That refusal is correct and it is also **unbounded**: a merged lane whose shells never
exit pins its build output forever. Two merged lanes currently pin **120 GB** this way — 113 GB
of it behind three shells aged 28 h, 39 h and 67 h.

## What investigating the real case changed about the design

Those shells are *not* idle. Each re-executes every 5–20 seconds and has a live parent Claude
CLI process. Each is also **incapable of ever finishing**:

```
until ! pgrep -f "scratchpad/mutate.py"; do sleep 20; done
```

`pgrep -f` matches command lines, and the waiting shell's own command line *contains that
pattern*. The shell matches itself, the predicate is permanently false, the loop spins forever.
A third waits on `grep -q "^EXIT="` over a file whose originating session was cleaned up and
which no longer exists. Both were confirmed directly: the awaited work has 0 real processes, and
the awaited file does not exist.

So these lanes look maximally alive by every naive metric — fresh child process, live parent,
constant re-execution — and are maximally dead in fact.

**The model therefore measures progress, not activity**, and reports activity-without-progress
as the specific pathology it is. Both non-terminating shapes are detected structurally, per
process. The `[b]racket` idiom (`pgrep -f "[m]utate.py"`) — the correct way to avoid
self-matching — is explicitly *not* flagged, because a false positive here proposes a healthy
process for termination.

## The existing lifecycle machinery is unwired — this is the root cause

The ops MCP already defines exactly the needed model:

```
sessions(id, repo, worktree, task_description, started_at, last_heartbeat)
register_session · heartbeat · release_session · list_sessions
```

with "active" meaning `last_heartbeat > now - 30 minutes`, bound to a worktree name.
**Nothing populates it.** `list_sessions` returns `[]`; `recent_sessions` reports no history.

An authoritative liveness signal exists and is not wired up, so the only observable evidence is
"a process has a cwd here" — which never expires. Wiring agent launch to
`register_session`/`heartbeat` would make most of this tool's heuristics unnecessary. That is
the real fix, and it is an agent-harness change rather than an ops script.

Until then, **absence of a heartbeat is treated as no signal, never as evidence of staleness.**

## States

| State | Retireable |
|---|---|
| `UNPINNED` no processes — the disk guard's business | n/a |
| `ACTIVE` build/test child, or progress within 90 min | no |
| `OPEN-PR` / `UNMERGED` | no |
| `PINNED-DIRTY` uncommitted source | no |
| `PINNED-UNKNOWN` unrecognised process, or unknown git state | no |
| `QUIESCENT` merged and quiet, inside the grace window | no |
| `STALE-CANDIDATE` merged, agent shells only, no work children, clean, quiet ≥ 8 h, oldest pin ≥ 4 h | **candidate only** |

Safety predicates dominate staleness: an unknown process outranks every other signal; dirty
state outranks staleness. Thresholds are in one config block, environment-overridable.

## Automatic retirement is deliberately NOT implemented

`--plan-retire` prints what termination *would* target and stops. Nothing here signals a
process. Three reasons, all of which would have to change first:

1. **No authoritative activity signal** — the session registry is empty, so liveness is inferred
   from filesystem and process heuristics. Good enough to report; not to kill.
2. **The parents are alive** — the pinning shells are children of running Claude CLI processes
   (1.7 and 2.8 days old) presumably blocked awaiting them. Terminating the child would likely
   *unblock* a stuck agent, plausibly a repair — but it reaches into another live session, and
   "plausibly" is not a safety argument.
3. **The detector is new** — it found the real pathology on its first run. That is not the same
   as being right.

Operator-approved procedure is documented in the README: review `--plan-retire`, `kill -TERM`
(never SIGKILL by default), re-audit, then let the disk guard apply its own rules.

## Contract with the disk guard

Deliberately implicit — neither tool calls the other, and the guard gains no process-control
powers. Verified end to end on a real merged lane (PR #2594) with a disposable synthetic pin:

```
1. pinned      -> guard: MERGED-BUT-PROCESS-PINNED  KEEP:process-pinned(1)
2. audit       -> lane:  QUIESCENT, retireable=no   (fresh pin; grace window protects it)
3. SIGTERM     -> pid exited cleanly
4. re-audit    -> guard: MERGED-INACTIVE  ELIGIBLE:routine   (its own rules, no coupling)
```

Step 2 doubles as a negative control: a *fresh* pin is correctly refused, not retired.

## Tests

**28/28 self-tests**, covering every case the design has to get right: merged+old shell with
recent activity → KEEP; without → candidate; open-PR with ancient shell → KEEP; unmerged →
KEEP; dirty → KEEP; unknown process → fail-safe; active build child → KEEP; young process →
never a candidate; inside the grace window → QUIESCENT; every non-candidate state asserted
non-retireable individually; and the three non-terminating detections including the bracket-idiom
negative control.

**Synthetic worktrees with real processes** for the destructive-adjacent cases; dry-run and
`--plan-retire` verified to terminate nothing (process count unchanged). No real agent process
was terminated at any point — the four pinning shells are still alive and were re-verified after
every step.

Three bugs were found by running it and are fixed here: basename path collision (a worktree
literally named `icn` claimed four other lanes' processes — same bug fixed in #2652), `/proc`
mtime misread as process start time (a 67 h shell reported as 1 h), and the auditor counting its
own probe processes as pins — itself an instance of the self-matching error it detects.

## Scheduling

None installed. Hourly-to-4-hourly would be ample; this is debt reporting, not monitoring, and
any automatic retirement threshold must stay far longer than the audit cadence.

## Non-goals

No process termination. No worktree deletion (`wt-clean` exists; ICN policy does not authorise
automatic worktree removal — the evidence that would make it safe is listed in the README as a
proposal). No changes to `icn-disk-guard` beyond the shared path-matching fix already in #2652.

### DIFF
--- a/ops/scripts/README-icn-lane-audit.md
+++ b/ops/scripts/README-icn-lane-audit.md
@@ -1,0 +1,123 @@
+# Agent-worktree lifecycle audit
+
+Reports lifecycle debt on the `icn-dev` development VM: merged lanes whose build output cannot
+be reclaimed because processes still hold the worktree.
+
+```bash
+ops/scripts/icn-lane-audit                 # human-readable
+ops/scripts/icn-lane-audit --json          # machine-readable
+ops/scripts/icn-lane-audit --plan-retire   # what retirement WOULD signal; acts on nothing
+ops/scripts/icn-lane-audit --self-test     # 28 policy assertions
+```
+
+Read-only. It never signals a process, deletes anything, or mutates git.
+
+## Why this exists
+
+`icn-disk-guard` refuses to reclaim a Cargo target while any process holds the worktree. That
+refusal is correct and it is also unbounded: a merged lane whose shells never exit pins its
+build output forever. Two merged lanes currently pin **120 GB** this way.
+
+## The finding that shaped the design
+
+The three shells pinning the merged #2644 lane (aged 28 h, 39 h and 67 h) are not idle. Each
+re-executes every 5–20 seconds and has a live parent. Each is also **incapable of ever
+finishing**:
+
+```
+until ! pgrep -f "scratchpad/mutate.py"; do sleep 20; done
+```
+
+`pgrep -f` matches command lines, and the waiting shell's own command line *contains that
+pattern*. The shell matches itself, the predicate is permanently false, the loop spins forever.
+A third waits on `grep -q "^EXIT=" <file>` for a file whose originating session was cleaned up
+and which no longer exists.
+
+So the lanes look maximally alive by every naive metric — fresh child process, live parent,
+recent process activity — and are maximally dead in fact.
+
+**The model therefore measures progress, not activity**, and treats activity-without-progress
+as the specific pathology it is. Both non-terminating shapes are detected structurally and
+reported per process.
+
+The `[b]racket` idiom (`pgrep -f "[m]utate.py"`) is the correct way to avoid self-matching and
+is explicitly *not* flagged — a false positive here would propose a healthy process for
+termination.
+
+## The existing lifecycle machinery is unwired
+
+The ICN ops MCP already defines exactly the model this needs:
+
+```
+sessions(id, repo, worktree, task_description, started_at, last_heartbeat)
+register_session · heartbeat · release_session · list_sessions
+```
+
+with "active" defined as `last_heartbeat > now - 30 minutes`, and a session bound to a worktree
+name. **Nothing populates it.** `list_sessions` returns `[]` and `recent_sessions` reports no
+history: agents never call `register_session` or `heartbeat`.
+
+That is the root cause of stale lanes. A designed, authoritative liveness signal exists and is
+not wired up, so the only observable evidence is "a process has a cwd here", which never
+expires. Wiring agent launch to `register_session`/`heartbeat` would make this tool's
+heuristics largely unnecessary — that is the real fix, and it is an agent-harness change rather
+than an ops-script one.
+
+Until then, **absence of a heartbeat is treated as no signal, never as evidence of staleness.**
+
+## States
+
+| State | Meaning | Retireable |
+|---|---|---|
+| `UNPINNED` | no processes; the disk guard's business | n/a |
+| `ACTIVE` | build/test child running, or progress within 90 min | no |
+| `OPEN-PR` | the lane still has a job | no |
+| `UNMERGED` | branch not merged | no |
+| `PINNED-DIRTY` | uncommitted source — unique state | no |
+| `PINNED-UNKNOWN` | unrecognised process, or unknown git state — fail safe | no |
+| `QUIESCENT` | merged and quiet, but inside the grace window | no |
+| `STALE-CANDIDATE` | merged, agent shells only, no work children, clean, quiet ≥ 8 h, oldest pin ≥ 4 h | **candidate only** |
+
+Safety predicates dominate staleness: an unknown process outranks every other signal, and dirty
+state outranks staleness. Thresholds live in one config block and are environment-overridable.
+
+## Automatic retirement is NOT enabled, and should not be yet
+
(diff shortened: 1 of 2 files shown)


fahertym:
Re-read from first principles after the agent-runtime bootstrap landed (#2671, `cd943e02`). **Not superseded. Keep.**

The supersession case was that `icn-wait` prevents the non-terminating wait shapes and the session registry reports lane liveness. Both are true, and neither covers what this PR covers. Today's session is the evidence, not a hypothetical:

**What the runtime now covers.** Before #2671, 44 of 50 worktrees could not register at all, so the registry was blind and every lane read as REGISTRY-UNAVAILABLE. That is fixed: a fresh lane registers automatically, and lane liveness for *registered sessions* is now authoritative.

**What it still does not cover, demonstrated today.** The residue actually found on this VM was invisible to the registry by construction, because none of it is a session:

- 16 orphaned processes under `task-2419-sprint-guard/icn` (`ppid=1`, age 1d22h) — a synthetic contention test whose `kill $LOADPIDS` never ran because its owning Claude activation died. They were holding load average at ~19.
- `sleep 999999` (`ppid=1`, age 3d21h) pinning `task-agent-runtime`.
- Together with merged-but-unretired lanes, **121 GB** (181G → 60G after retirement).

I found all of it by hand-walking `/proc/*/cwd` and correlating against worktrees, PR state and dirty/unpushed status. That is precisely what `icn-lane-audit` automates.

**Suggested narrowing** for whoever picks this up, so it does not become a second lifecycle model:

1. Take registered-session facts from the runtime and label them as such — do not re-derive liveness where `icn-agent-session classify` already answers.
2. Scope the PR's own analysis to what the runtime cannot see: processes with no owning session, orphans (`ppid=1`), and lanes whose disk cost is not justified by their PR state.
3. Keep every non-runtime signal explicitly labelled a heuristic, and keep it read-only. The `[b]racket` exemption stays right.

Reporting only, no process-killing authority — that judgement stayed with a human today and should keep doing so.

fahertym:
## Disposition: **KEEP_DRAFT** — reconciled against `main` at `3c1e4f863` (2026-09-06)

**Not superseded, and the old-numbers concern does not apply here.** I checked specifically for the failure mode where a dev-host PR bakes in historical facts:

- Paths are parameterized through `ICN_DEV_ROOT` / `$HOME/icn-dev` — no hardcoded host addresses.
- Thresholds are named configuration variables, not literals buried in logic.
- The numeric values that do appear in the diff are **test fixtures**, not asserted facts about any current host.

So this is not carrying a stale 2026-08 snapshot of the VM forward as truth.

**Nor is the solution shape superseded.** Newer session infrastructure — `.claude/hooks/session-lifecycle.sh` and the `ops/mcp` session tools — covers *session* lifecycle: registration, heartbeat, release. This pair covers *worktree and disk* debt. Non-overlapping; neither makes the other redundant.

**What actually blocks these is not code — it is a placement decision.** Both are dev-VM operations tooling living in the product repository, and both have **zero reviews**. The open question a maintainer must answer is whether `ops/` in this repo is the right home for icn-dev host hygiene, or whether it belongs in the operations repo.

**#2652 and #2653 should be decided together** — they are the same shape, the same scope question, and the same author, and splitting the decision would leave one orphaned.

They are read-only, valid, uncontested, and cost nothing to leave as drafts. Conflict risk **LOW** (zero competing commits since merge-base `abdfab79`). Keeping them as drafts pending that placement call is the honest state — not a rebase problem, and not a close.

Issue #2678, opened 2026-08-28
Follow-up ledger: PR #2677 / #2641 governance vote-principal review


fahertym:
> **Revalidated 2026-09-06 against `main` at `3c1e4f863`.** Original items 1, 4 and 6 are **discharged by I7** (#2686) and item 2 was resolved in #2677; all four are removed. Original numbering is retained below so the audit trail in the comment stays readable.

Deferred review observations from PR #2677 / #2641 (governance vote-principal work). Three of the four surviving items are **live §7.5 membership/vote migration debt** and must not be closed by hardening that does not migrate.

## Item 3 — no repair path for a principal that already holds several rows

Pre-#2641 stores can hold several vote rows for one principal under different spellings. Nothing repairs them. The gate document is explicit — *"No merge rule, no survivor, no repair"* (`docs/architecture/n2-a-migration-gate.md:1880`) — and §11.4 still lists bare `vote:` rows as **uncovered** by any deferred namespace, so they are not even reported by the startup gate.

**This is §7.5 debt.** Choosing between two conflicting historical acts is the §7.5 gate's business. Do not invent a survivor rule here.

## Item 5 — `GovernanceStore` has no liveness statement, and the cleanup task is unwired

`pub trait GovernanceStore` (`icn-governance/src/store.rs:53-54`) documents only *"Trait for governance storage operations"* — nothing about what an implementation must guarantee about row liveness. `ProposalCleanupTask::new` appears only in a doc-comment and in tests (`proposal_cleanup.rs:20,582,606,633,751`); no production path constructs it.

An unwired cleanup task plus an unstated liveness contract means no one can say whether stale proposal rows are expected to disappear.

## Item 7 — the decision receipt is built from the unreduced vote set

`GovernanceDecisionReceipt::new(..., tally.clone(), &proposal_votes)` (`apps/governance/src/manager.rs:4409-4415`) receives `proposal_votes`, the raw/eligibility-filtered row set assembled at `:4294-4298` — **never** the principal-reduced set the tally was computed from.

So the receipt's evidence and its tally can disagree for a principal holding several spelling-keyed rows. **This is a semantics decision, not a bug fix:** does a receipt attest the rows observed, or the effective votes counted? Both are defensible; the code currently does not say which it means. Decide it explicitly before N2-A closes.

## Item 8 — the multi-row refusal is load-bearing and must survive

[...]

Issue #2639, opened 2026-08-19
fix(website): rendered audit fails on a real +6px overflow of the ladder-here pill (Chrome diagnosis stale); also pin CHROME_BIN and stop discarding launch diagnostics


fahertym:
## Problem

`Website CI` → **Build and verify** → *"Rendered layout and accessibility audit"* began failing on **2026-08-19 between 18:23 and 18:50 UTC** with no corresponding code change:

```
Error: Chrome did not expose a DevTools endpoint
    at cdpTarget (website/scripts/audit-layout.mjs:233:9)
```

The failure is a **browser launch** failure, not a site failure. Every content-dependent step in the same job passed — *Build site*, *Type check*, *Checker unit tests*, *Public state projection matches canonical state*, *Documentation publication boundary*, *Internal links and compatibility redirects*, *Readiness overclaim*. Only step 13 failed.

## Evidence that it is environmental

Identical `website/` tree, same workflow, same day:

| Time (UTC) | Branch | SHA | Result |
|---|---|---|---|
| 17:49 | main | `798c8d54` | success |
| 18:11 | fix/2634-… | `2f894709` | success |
| 18:17 | fix/2634-… | `5bed8880` | success |
| 18:23 | docs/n2-a0-… | `bd85cd0d` | success |
| **18:50** | fix/2634-… | `5a4f38ff` | **failure** |
| **19:42** | fix/2634-… | `5a4f38ff` (re-run) | **failure** |

`5a4f38ff`'s `website/` tree is byte-identical to `798c8d54` (`git diff --quiet origin/main..HEAD -- website/` is clean), and the PR touches no website file. Code that did not change cannot explain a status that did. Re-running the job once did not clear it, so this is a persistent environment regression, not a transient flake.

## Root cause of the *fragility*

`.github/workflows/website-ci.yml` deliberately takes whatever Chrome the runner image ships:

> `audit-layout.mjs` finds Chrome on PATH (ubuntu runners ship it) and reports exactly what it tried if it cannot. **No `CHROME_BIN` pin, and no browser download.**

That makes the audit a function of the `ubuntu-latest` image rather than of the repository, so an image-side Chrome change breaks CI on every PR touching `website/**` or `docs/**` with nothing in the diff to explain it.

## Secondary defect: the failure reports nothing useful

[...]

PR #2721, opened 2026-09-06
fix(ci): repair the benchmark regression count; docs(identity): correct a §7.5 claim I7 made false

fahertym:
Two unrelated, independently-verified truth repairs found during a repository reconciliation pass. Both are isolated and neither touches the protected N2-A M4d lane (#2716 — `icnctl`, `icn-store`, `n2-a-migration-gate.md`).

## 1. `fix(ci)` — benchmark regression count evaluated a two-line string

`grep -c` prints `0` on no match **and** exits 1, so `|| echo "0"` fired *on top of* grep's own output and `REGRESSION_COUNT` became `"0\n0"`. The following `[ "$REGRESSION_COUNT" -gt 0 ]` then failed with `integer expression expected` instead of comparing, and the `if` fell through to the no-regression branch.

Clean runs reached the correct verdict **by way of a shell error rather than a comparison**, logging a spurious error each time. Two-or-more regressions still compared correctly — so the reporting worked in the case that mattered and hid the defect in the case that did not.

**Verified both branches directly** before and after:

```
before:  no match -> "0\n0"  -> `[: integer expression expected`
after:   no match -> "0"     -> else branch, no error
after:   2 matches -> "2"    -> regression branch
```

Workflow-only. No benchmark, threshold, or Rust change.

## 2. `docs(identity)` — a §7.5 claim that I7 made false

`docs/architecture/IDENTITY_SEMANTICS.md` is the registered owner of the `identity_semantics` domain and is marked `Status: normative` / `Canonical: yes`.

§7.5 closed by asserting that membership lists are compared by DID spelling and that *"under a static list an alias spelling is refused at the membership gate, which is fail-closed"*. That described pre-I7 behaviour.

Since **#2686** (`0defbde51`), `Did::PartialEq` compares decoded identifier bytes, so `MembershipSource::contains_static` — a `Vec<Did>` tested with `contains` (`icn/crates/icn-governance/src/membership.rs:46`) — now **admits** any accepted spelling of a listed member. **The document claimed a fail-closed refusal where the code admits**, which is the unsafe direction for a reader reasoning about the gate's posture.

The replacement text separates the two things §7.5 was conflating: in-memory membership *comparison* (which moved with I7) and the persisted membership/vote *keys* this gate governs (which have **not** been re-keyed).

**This is synchronization with landed code, not a semantic or governance change.** The gate itself, its four prerequisites, and the *"no persisted vote or membership key has been re-keyed"* statement are all unchanged. Surfaced while revalidating #2678.

## Verification

- `python3 docs/scripts/doc_control_check.py` → **passes** (971 markdown files; the 64 enforcement warnings are pre-existing and name other documents).
- `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/benchmark.yml'))"` → valid.
- Both `grep` branches exercised directly (output above).
- No Rust source changed, so no `cargo` gate is in scope for this diff.

## Non-claims

- Neither change fixes any advisory, benchmark regression, or migration.
- The §7.5 correction advances **nothing** in N2-A. It makes an existing document true; #2627 stays open and §7.5 remains a separate hard gate.

Refs #2627 #2678 #2686.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -145,7 +145,7 @@ jobs:
 
           # Count only Criterion's explicit regression verdict. Do not match confidence-interval
           # lines like `change: [-0.5% -0.2% +0.03%]` — the positive upper bound is not a regression.
-          REGRESSION_COUNT=$(grep -ciF "Performance has regressed" bench_output.txt || echo "0")
+          REGRESSION_COUNT=$(grep -ciF "Performance has regressed" bench_output.txt || true)
 
           if [ "$REGRESSION_COUNT" -gt 0 ]; then
             echo "has_regressions=true" >> $GITHUB_OUTPUT
--- a/docs/architecture/IDENTITY_SEMANTICS.md
+++ b/docs/architecture/IDENTITY_SEMANTICS.md
@@ -1,7 +1,7 @@
 ---
 Status: normative
 Canonical: yes
-Last Reviewed: 2026-08-18
+Last Reviewed: 2026-09-06
 ---
 
 # Identity Semantics — the N2 semantic contract
@@ -499,9 +499,21 @@ them all at once.
 
 The other three prerequisites — migration ordering, alias/transition
 recognition, final cutover — remain undesigned, and **no persisted vote or
-membership key has been re-keyed**. Membership lists are still compared by DID
-spelling; under a static list an alias spelling is refused at the membership
-gate, which is fail-closed but is not yet alias recognition.
+membership key has been re-keyed**.
+
+In-memory membership *comparison* is a separate matter from the persisted
+*keys* this gate governs, and it has moved. Since I7 landed (#2686), `Did`
+equality is principal equality, so a static membership list — a `Vec<Did>`
+tested with `contains` — now admits **any** accepted spelling of a listed
+member. A previous revision of this section stated that an alias spelling is
+refused at the membership gate and called that fail-closed; that described
+pre-I7 behaviour and is no longer true.
+
+This does not narrow the gate. Admitting an alias of a genuine member is the
+correct reading of principal equality, and it is why the three remaining
+prerequisites still matter: recognition at the comparison boundary is not
+migration of the persisted keyspace, and nothing above re-keys a stored
+membership or vote row.
 
 ---
 


fahertym:
# Merge-decision packet — #2721

**Status: ready for a human merge decision. Not merged — no authorization present in this session.**

## Exact identifiers

| | |
|---|---|
| Head | **`9860d323fd056ac47e2c67033c65456b1a61f8af`** |
| Base | `main` = **`3c1e4f863`** — **unmoved since this PR was opened, so no rebase is needed** (`strict: true` satisfied) |
| Diff | 2 files, 2 commits, **+17 / −5**. No Rust source. |

## Required checks — 11/11 SUCCESS

All eleven live required contexts are green at this exact head. The single red, `Build and verify`, is **non-required** and is the separately-owned **#2639** `ReadingLadder.astro` +6px overflow — red on `main` itself and inherited, not caused here. `mergeStateStatus: UNSTABLE` reflects only that.

**Review:** 1 Copilot review (COMMENTED), **0 unresolved threads**.

## Re-verified against current `main` this session

**1 — the benchmark fix still solves the reproduced bug.** `.github/workflows/benchmark.yml:148` now reads `|| true`. Both branches exercised directly just now:

```
no match  -> "0"  -> comparison takes the else branch, no error
2 matches -> "2"  -> comparison takes the regression branch
```

Before the fix, `grep -c` printed `0` **and** exited 1, so `|| echo "0"` fired on top of it and `REGRESSION_COUNT` became the two-line string `"0\n0"`; `[ … -gt 0 ]` then failed with `integer expression expected` instead of comparing. Clean runs reached the right verdict **via a shell error rather than a comparison**.

**2 — the §7.5 text remains synchronized with actual `MembershipSource` behaviour.** Live code:

```rust
// icn/crates/icn-governance/src/membership.rs:46
MembershipSource::StaticList(members) => members.contains(did),
```

`members` is `Vec<Did>`, so `contains` uses `Did::PartialEq`, which since #2686 compares decoded identifier bytes. The corrected prose says exactly that: a static list *"now admits **any** accepted spelling of a listed member"*. The previous text claimed such a spelling *"is refused at the membership gate, which is fail-closed"* — wrong, and wrong in the direction that would mislead a reader about the gate's safety posture.

`python3 docs/scripts/doc_control_check.py` passes (971 files; its 64 enforcement warnings are pre-existing and name other documents).

## Acceptance contract

Make two existing truth surfaces true: the benchmark workflow must compare an integer, and `IDENTITY_SEMANTICS.md` §7.5 must not assert behaviour that I7 inverted.

**Non-goals:** no benchmark threshold or policy change · no advisory remediation · no website change · **no N2-A migration or cutover of any kind**.

## What merging this does and does not mean

- ✅ Removes a spurious shell error from every clean benchmark run, and makes the comparison real.
- ✅ Removes a false safety claim from a `Canonical: yes` registered truth owner.
- ❌ **The §7.5 edit is synchronization only. It advances no migration gate**, does not narrow §7.5, and leaves *"no persisted vote or membership key has been re-keyed"* untouched.
- ❌ Does not advance **#2627**, and does not fix **#2639** or **#2579**.

## Decision required

**Merge authorization for #2721 at exactly `9860d323fd056ac47e2c67033c65456b1a61f8af`.** Squash, no `--admin`.

fahertym:
## Re-verified after `main` moved — **new exact head**

`main` advanced to **`214ed2d948da`** when #2716 merged, and this branch was updated (merge commit **`f7f7bf19a490ad829f3a9b9640aa43f04d6abaa6`**). **The earlier packet's head `9860d323…` is stale and must not be authorized.**

| | |
|---|---|
| **Exact head** | **`f7f7bf19a490ad829f3a9b9640aa43f04d6abaa6`** |
| Base | `main` = `214ed2d948da…` — merge-base **equals** `origin/main`, `strict: true` satisfied |
| Required CI | **11/11 SUCCESS** on this head |
| Threads | 0 unresolved |

**Content is unchanged.** Delta against the new `main` is still exactly the two files:

```
.github/workflows/benchmark.yml           2 +-
docs/architecture/IDENTITY_SEMANTICS.md  20 ++++++++++++++++----
2 files changed, 17 insertions(+), 5 deletions(-)
```

The only red is **`Build and verify`**, which is non-required and is the separately-owned #2639 `ReadingLadder.astro` overflow — red on `main` itself, inherited, not caused here.

### Both claims re-verified against the new base

- **Benchmark fix** — `benchmark.yml:148` reads `|| true`; both branches exercised: no match → `"0"` → else branch, no error; 2 matches → `"2"` → regression branch.
- **§7.5 synchronization** — still matches live code. `MembershipSource::contains_static` is `members.contains(did)` over `Vec<Did>` (`icn-governance/src/membership.rs:46`), so principal equality admits any accepted spelling, which is what the corrected prose states.

### Non-claims, restated

**The §7.5 edit is synchronization with landed code. It advances no migration gate**, does not narrow §7.5, and leaves *"no persisted vote or membership key has been re-keyed"* untouched. It does not touch #2627, which stays OPEN.

### Decision required

**Merge authorization for #2721 at exactly `f7f7bf19a490ad829f3a9b9640aa43f04d6abaa6`.** Squash, no `--admin`. Not merged — no authorization is present in this session.

PR #2762, opened 2026-09-10
ci: bound every job, stop duplicate runs, and drop a fragile required-check dependency

fahertym:
## Delivery

- **Delivery lane**: FAST (CI configuration only; no product code)
- **Acceptance contract**: no CI job can run unbounded, no required check depends on a job it does not read, and no workflow runs the same work twice for one push.
- **Explicit non-goals**: does not change what any check verifies, does not add/remove/rename a required check, does not touch `security-audit.yml`, `claude.yml` or `opencode.yml` (changed in #2760 / #2761).

## 1. Four required checks depended on a job they never read

`secrets-check`, `meaning-firewall`, `kernel-deps`, `firewall-contract`, `compliance-linter` and `readiness-overclaim` each declared `needs: [changes]` and referenced `needs.changes.outputs` **nowhere**. Four are required by branch protection.

That dependency was not free:

- each waited on a checkout + a `dorny/paths-filter` call to the PR API before starting;
- each inherited `changes`'s failure mode — **one paths-filter API blip skipped four required checks at once**, leaving the PR unmergeable until someone re-ran CI by hand.

`fmt` already had no such dependency, which is the correct shape.

Verified mechanically by parsing the workflow before and after:

```
origin/main: ['secrets-check: vestigial needs', 'meaning-firewall: vestigial needs',
              'kernel-deps: vestigial needs', 'firewall-contract: vestigial needs',
              'compliance-linter: vestigial needs', 'readiness-overclaim: vestigial needs']
this branch: clean
```

The same check confirms **no job that reads `changes.outputs` lost its `needs`** — including `pilot-provenance`, which declares `needs: [changes, test]`.

## 2. No job in the repository had a timeout

40 jobs across 16 workflows now carry `timeout-minutes`; a hung step previously burned the 6h default while holding a required check pending.

Two worth naming:

- **`accessibility`** (required) runs `npx playwright install --with-deps chromium` — an `apt-get` under the hood. A stalled mirror held a required check pending for six hours with no signal.
- **`backup-validation`** runs `icnctl id init`. `rpassword` reads `/dev/tty`, never stdin, so a regression in the env-passphrase path **hangs** rather than fails.

## 3. Work done twice, or thrown away

| Problem | Fix |
|---|---|
| `docs-freshness` + `evaluator-package` had `push:` with a paths filter but **no branches filter** — every feature-branch push ran them, then `pull_request` ran them again | `push: branches: [main]` |
| `agent-drift-check` (required), `api-types`, `docs-freshness`, `evaluator-package` had no concurrency group — superseded pushes kept running | concurrency group per PR; cancels on PR, never on main |
| `api-types` cached `./target` but built in `icn/` — cache could **never** hit, cold `icnctl --release` build every run | `cache-workspaces: "icn -> target"` |
| `icn-binaries` (release icnd + icnctl + icn-console) uploaded on every PR and push, kept 90 days | `retention-days: 5` |
| a11y reports kept 90 days | 7 |
| benchmark baseline kept 90 days | 14 |
| `website-ci` on upload-artifact@v4, `evaluator-package` on checkout@v4, everything else @v7/@v6 | aligned |

Release artifacts keep the 90-day retention — those are releases.

## Verification

- All 21 workflow files parse (`yaml.safe_load`).
- `needs`/`outputs` consistency proven by parse, before and after (output above).
- Trigger shapes confirmed post-edit: `docs-freshness` and `evaluator-package` are `push: branches: [main]` with `pull_request` still filtered by `paths`.
- Every job in every workflow now has `timeout-minutes` (verified by parse).
- Branch-protection contexts untouched.

## Invariants

- **determinism** — strengthened on CI: a required check no longer depends on an unrelated PR-API call.
- adversarial-by-default, canonical encodings, no-panics, kernel/app boundaries — not touched (CI configuration only).

## Note on merge order

Touches `ci.yml`, as does #2760, in different regions (that PR edits step bodies in the `security` and `sdk` jobs; this one edits job headers). Whichever lands second may need a trivial rebase.

### DIFF
--- a/.github/workflows/agent-drift-check.yml
+++ b/.github/workflows/agent-drift-check.yml
@@ -114,6 +114,13 @@ jobs:
       - name: Sprint-state currency + SessionStart source guard tests (Refs icn#2634)
         run: python3 scripts/tests/test_sprint_state_invariants.py
 
+      # A job timeout below its own steps' budgets cancels the step before it
+      # can use the time it was given, and reports a timeout instead of the real
+      # result. This job runs on every pull request with no paths filter, so it
+      # is the surface that sees workflow edits.
+      - name: Workflow job timeouts exceed their nested step budgets
+        run: python3 scripts/tests/test_workflow_timeout_budgets.py
+
       # icn#2651: the canonical merge policy must not name a field that cannot hold the value it
       # is compared against, and must not duplicate values it owns. Both got past every other gate
       # because prose is not type-checked. Structured validation only — no Markdown is parsed.
--- a/.github/workflows/agent-drift-check.yml
+++ b/.github/workflows/agent-drift-check.yml
@@ -57,10 +57,17 @@ on:
 permissions:
   contents: read
 
+# A superseded push to a PR branch should not keep burning a required check.
+# Never cancels on push to main.
+concurrency:
+  group: agent-drift-${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
 jobs:
   drift-check:
     name: Agent Tooling Drift Check
     runs-on: ubuntu-latest
+    timeout-minutes: 20
     steps:
       - uses: actions/checkout@v6
 
--- a/.github/workflows/api-types.yml
+++ b/.github/workflows/api-types.yml
@@ -11,10 +11,15 @@ on:
 permissions:
   contents: read
 
+concurrency:
+  group: api-types-${{ github.event.pull_request.number || github.sha }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
 jobs:
   check-api-types:
     name: Check API Types Drift
     runs-on: ubuntu-latest
+    timeout-minutes: 45
     steps:
       - uses: actions/checkout@v6
 
@@ -25,6 +30,11 @@ jobs:
         uses: actions-rust-lang/setup-rust-toolchain@v1
         with:
           toolchain: 1.95.0
+          # The Cargo workspace is icn/, not the repo root. The action's default
+          # cache-workspaces is ".", so the cached path (./target) was never the
+          # built path (icn/target) and this job cold-built icnctl --release on
+          # every run. Matches the `workspaces:` used elsewhere in the repo.
+          cache-workspaces: "icn -> target"
 
       - name: Setup Node.js
         uses: actions/setup-node@v6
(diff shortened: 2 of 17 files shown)


fahertym:
## All three timeout findings fixed — 59636f942

All three were the same defect: a job `timeout-minutes` set **below** the step budgets already declared inside that job, which cancels the step before it can use the time it was given and reports a timeout instead of the real result.

| job | was | now | why |
|---|---|---|---|
| `benchmark.yml:benchmark` | 75 | **105** | one 90-minute step + setup |
| `benchmark.yml:compare` | 75 | **140** | two sequential 60-minute passes + two checkouts + toolchain |
| `fuzz.yml:fuzz` | 60 | 60 (input bounded instead) | see below |

`fuzz` is a different shape. Its budget is correct for the 300s default and wrong only for a dispatch asking for more than the job can honour — three targets run **sequentially**, each for the full duration. Raising the job budget to cover an unbounded input would not fix that, so the **input is bounded against the budget**: `FUZZ_MAX_DURATION=900` gives 45 minutes of fuzzing inside the 60-minute job, and a larger request is refused up front with a message saying why. A request the job cannot honour is a bad request, not a failure.

While in that validator: it claimed "positive integer" but `^[0-9]+$` accepts `0`. The lower bound is now checked rather than implied.

## The enabling mechanism, not just the three instances

These two numbers are written in different places, by different people, at different times, and **nothing related them**. That is why one commit introduced three instances, and why review had to do the arithmetic by hand three times.

`scripts/tests/test_workflow_timeout_budgets.py` now executes that arithmetic: for every job declaring `timeout-minutes`, it sums its steps' budgets — they run sequentially, so they add — and requires the job budget to exceed the sum plus setup headroom.

Verified in **both** directions:

```
$ sed -i 's/timeout-minutes: 105/timeout-minutes: 75/' .github/workflows/benchmark.yml
$ python3 scripts/tests/test_workflow_timeout_budgets.py
test_workflow_timeout_budgets: FAIL
  - benchmark.yml:benchmark: job timeout-minutes=75 but its steps declare [90] (sum 90)
    and run sequentially. The job would cancel a step before that step's own budget
    expired. Need at least 95 (sum + 5 setup headroom).
exit=1

$ # restored
$ python3 scripts/tests/test_workflow_timeout_budgets.py
test_workflow_timeout_budgets: clean (2 job(s) with nested step budgets verified)
exit=0
```

Wired into `agent-drift-check.yml`, which runs on **every** pull request with no paths filter and is therefore the surface that sees workflow edits.

This addition is deliberately inside this PR's acceptance contract rather than deferred: the contract is "no CI job can run unbounded", and a bound that cancels its own steps does not satisfy it. Scope did not otherwise widen.

PR #2763, opened 2026-09-10
fix(deps): take the wasmtime patch for the filesystem sandbox escape (RUSTSEC-2026-0269)

fahertym:
## Delivery

- **Delivery lane**: FAST (single lockfile bump inside an already-declared semver range)
- **Acceptance contract**: `cargo audit` no longer reports wasmtime, and the WASM execution path still builds and passes its tests.
- **Explicit non-goals**: no manifest edit, no wasmtime major upgrade, no change to `deny.toml`, no suppression of any advisory.

## Problem

`wasmtime` 36.0.13 is affected by **RUSTSEC-2026-0269** — a filesystem sandbox escape when paths or symlinks carry trailing slashes, **CVSS 8.8 (high)**.

This is not an incidental transitive dependency. ICN executes WASM modules through `icn-compute` (`wasmtime = { version = "36.0.8", optional = true }`, feature `wasm`), and `icn-baseline-lock` links it too. wasmtime's filesystem sandbox is precisely the boundary ICN relies on to contain untrusted module code, so a sandbox escape is a containment failure in the adversarial-by-default sense, not a dependency hygiene nit.

It has been reported by the Weekly Security Audit every week since **2026-08-17** and was never actioned — because, as #2760 shows, the audit's failure also suppressed the adjacent cargo-deny gates, so the whole job read as one chronic red rather than as a specific, fixable finding.

## Change

The advisory's fixed versions are `>=24.0.13 <25`, `>=36.0.14 <37`, `>=46.0.3 <47`, `>=47.0.4`.

The workspace already declares `wasmtime = "36.0.8"`, so **36.0.14 is inside the range the manifest already requests**. This is a patch-level lockfile bump, not an upgrade:

```
wasmtime 36.0.13 -> 36.0.14   (+ 18 internal companion crates, same bump)
```

`Cargo.lock` only. Nothing else in the dependency graph moved — `cargo update -p wasmtime` reported 293 unchanged dependencies still behind latest, so this is the *smallest* remediation that clears the advisory, not an opportunistic refresh.

## Verification

- `cargo audit` (cargo-audit 0.22.1) against the updated lockfile: **wasmtime no longer appears** in the vulnerability output. Before the bump it was reported with `Severity: 8.8 (high)`.
- `cargo test -p icn-compute --features wasm` — result recorded in the PR thread.
- No manifest changed, so no downstream crate's declared requirements shift.

## The remaining advisory, and why it is not suppressed here

`h2` 0.3.27 (**RUSTSEC-2026-0258**, unbounded empty DATA frames) still appears. It is *not* fixed in this PR, deliberately:

- It enters solely through `actix-http` → `actix-web`/`actix-files` → `icn-gateway`.
- **actix-http 3.13.1 — the newest 3.x, and exactly what Dependabot proposes in #2151 — still requires `h2 ^0.3.27`.** So #2151 does not resolve this advisory, and no in-range bump exists. The fixed version is `>=0.4.16`, which the actix-web 4.x line does not use.
- Reachability: `h2` is an *optional* dependency of `actix-http`, enabled by actix-web's default `http2` feature. actix-web 4 negotiates HTTP/2 only via ALPN over TLS, and `icn-gateway` never binds TLS (no `bind_rustls` / `bind_openssl` anywhere in `crates/icn-gateway/src` or `bins/`). On current evidence **the vulnerable code path is present in the graph but not reachable in ICN's deployment.**

That makes h2 a documented-exception candidate rather than a code fix — but every existing entry in `deny.toml`'s `ignore` list is an *unmaintained* warning, never a live vulnerability. Adding the first real CVE to that list is a maintainer decision about the security posture, not a maintenance action, so it is written up in #2579 instead of being applied silently here.

## Invariants

- **adversarial-by-default** — directly strengthened: restores the integrity of the sandbox containing untrusted WASM modules.
- determinism, canonical encodings, no-panics, kernel/app boundaries — unaffected; no source change, no API change, no wire-format change.

Refs #2579

### DIFF
--- a/icn/Cargo.lock
+++ b/icn/Cargo.lock
@@ -1311,36 +1311,36 @@ dependencies = [
 
 [[package]]
 name = "cranelift-assembler-x64"
-version = "0.123.13"
+version = "0.123.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5f8e1303ae2128891cb59691a74de4547dd208bc8511a2f287cca2b93bb3c728"
+checksum = "6835dba958b2ab7ab523e7e99296e0524317f60430a00cf5850562ef78ea7001"
 dependencies = [
  "cranelift-assembler-x64-meta",
 ]
 
 [[package]]
 name = "cranelift-assembler-x64-meta"
-version = "0.123.13"
+version = "0.123.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "58b2740a5936332028d9a1e8f29a199de2fd386e426d44da5fea70cf8e3f8e75"
+checksum = "0b6e4ce8ee6d899381fbdd9e6561336c651189d46cecaeee09b29e8d80aa786e"
 dependencies = [
  "cranelift-srcgen",
 ]
 
 [[package]]
 name = "cranelift-bforest"
-version = "0.123.13"
+version = "0.123.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "37fd128d3629fb105e433bda09744c5a2959cd1da04617a455c4cc17dff1ebef"
+checksum = "0cb6d37015df7ea4b60450c1229ad5f5819a1fb27434b063f8e6216dfbd0c42a"
 dependencies = [
  "cranelift-entity",
 ]
 
 [[package]]
 name = "cranelift-bitset"
-version = "0.123.13"
+version = "0.123.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "24a88f6d5a6cf6fcbc6386415d48948094721dfa4585d6615938b11ca938f20a"
+checksum = "986bea0b0858b55192782120032ce9c15943fa073f186f6e479653c59e62c329"
 dependencies = [
  "serde",
  "serde_derive",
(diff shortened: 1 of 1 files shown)


fahertym:
## Local verification

`cargo test -p icn-compute --features wasm --no-fail-fast` against wasmtime 36.0.14:

```
17 test suites, 0 failures
  366 passed  (lib)
   20 passed
   19 passed
   13 passed
    4 passed
    4 passed
  … 0 failed; 0 ignored across all suites
```

`cargo audit` (cargo-audit 0.22.1) against the updated lockfile — **wasmtime no longer appears**. Before the bump it was reported as:

```
Crate:     wasmtime
Title:     Filesystem sandbox escape when paths or symlinks contain trailing slashes
ID:        RUSTSEC-2026-0269
Severity:  8.8 (high)
Solution:  Upgrade to >=24.0.13, <25.0.0 OR >=36.0.14, <37.0.0 OR >=46.0.3, <47.0.0 OR >=47.0.4
```

Remaining `cargo audit` output is unchanged from main: `h2` 0.3.27 (see the PR description for why it is not addressed here) plus the pre-existing `unmaintained`/`unsound`/`yanked` warnings already carried in `deny.toml`.

PR #2769, opened 2026-09-10
fix(deps): clear the non-breaking npm advisories in three workspaces

fahertym:
## Delivery

- **Delivery lane**: FAST (lockfiles only; no package.json, no source)
- **Acceptance contract**: every npm advisory with an in-range fix is cleared in the workspaces where one exists, and each of those workspaces still builds and tests.
- **Explicit non-goals**: no semver-major upgrade, no declared dependency range moved, no source change, no workspace touched where the only remedy is a migration.

## Change

`npm audit fix` **without `--force`** — semver-compatible resolutions only — in the three workspaces where every remaining advisory had an in-range fix.

| workspace | before | after |
|---|---|---|
| `sdk/typescript` | 8 | **0** |
| `web/pilot-ui` | 6 | **0** |
| `sdk/react-native` | 14 | **4** |

Cleared: `js-yaml` (GHSA-5p4m-2wfm-xmqj), `browserslist` (GHSA-73wf-gq98-2v4g), `fast-uri`, `brace-expansion`, `nanoid`, and the `baseline-browser-mapping` chain.

**These are the same advisories behind roughly a dozen individually-open Dependabot PRs.** This clears them across three lockfiles in one change — which matters because `main` is strict-protected with no merge queue, so each of those PRs would otherwise have cost its own branch update and full CI run.

## What is deliberately left

**`sdk/react-native`, 4 remaining** — `image-size`, `metro`, `metro-config`, `metro-transform-worker`. npm reports them fixable, but only by moving Metro, which is a React Native toolchain upgrade, not a lockfile bump.

**`ops/mcp`, 11 remaining — and this one is a defect, not a decision.** `npm audit fix` cannot run there at all:

```
npm error Cannot read properties of null (reading 'edgesOut')
```

It aborts identically on a real run and on `--dry-run`. So the agent-facing MCP server's 11 advisories (including high-severity `fast-uri`, `hono`, `ip-address`, `nanoid`, `postcss`) are **unreachable by the normal remediation path** until its lockfile is rebuilt. That is a bigger, riskier change than this PR — rebuilding a lockfile can move many versions at once — so it is filed separately rather than forced through here.

**`website`, `examples/mobile-app`, `sdk/react-native/examples/CoopWallet`** — remaining advisories need semver-major upgrades (`astro` 7.x, Expo SDK). Those are migrations. The `astro` one is *critical* (GHSA-26w7-cxv4-gfx2) and needs its own deliberate PR.

## Verification

Run in each changed workspace **after** the lockfile change:

| workspace | commands | result |
|---|---|---|
| `sdk/typescript` | `npm run build`, `npm test`, `npm run typecheck:examples`, `python3 check_sdk_doc_terms.py --sdk-root .` | all exit 0 |
| `web/pilot-ui` | `npm test` | exit 0 |
| `sdk/react-native` | `npm run build`, `npm test` | exit 0 |

`check_sdk_doc_terms.py` reported: *"OK: SDK docs/examples use canonical ICN ledger primitives (checked against 105 exported methods in src/index.ts)."* — the required `TypeScript SDK` gate is satisfied locally.

Only `package-lock.json` files are in the diff; `git status` showed no `package.json` modified, so no declared range moved and no downstream consumer's resolution constraints change.

## Invariants

Not applicable to protocol behaviour — JavaScript dependency lockfiles only. No Rust, no wire format, no kernel/app boundary.

### DIFF
--- a/sdk/react-native/package-lock.json
+++ b/sdk/react-native/package-lock.json
@@ -40,14 +40,14 @@
       "license": "MIT OR Apache-2.0",
       "devDependencies": {
         "@types/jest": "^30.0.0",
-        "@types/node": "^25.9.1",
+        "@types/node": "^25.9.2",
         "@types/ws": "^8.5.10",
-        "@typescript-eslint/eslint-plugin": "^8.59.4",
+        "@typescript-eslint/eslint-plugin": "^8.61.0",
         "@typescript-eslint/parser": "^8.56.0",
-        "eslint": "^10.4.0",
+        "eslint": "^10.4.1",
         "jest": "^30.4.2",
         "openapi-typescript": "^7.13.0",
-        "ts-jest": "^29.4.10",
+        "ts-jest": "^29.4.11",
         "ts-node": "^10.9.2",
         "typescript": "^5.3.0"
       },
@@ -56,11 +56,13 @@
       }
     },
     "../typescript/node_modules/@babel/code-frame": {
-      "version": "7.27.1",
+      "version": "7.29.7",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.29.7.tgz",
+      "integrity": "sha512-Aup7aUOfpbAUg2ROOJN6Iw5f9DMBlzu0mIkm/malLQFN/YQgO48wCj0Kxa3sEHJvPVFg7siR+qRInwXd2qhQKw==",
       "dev": true,
       "license": "MIT",
       "dependencies": {
-        "@babel/helper-validator-identifier": "^7.27.1",
+        "@babel/helper-validator-identifier": "^7.29.7",
         "js-tokens": "^4.0.0",
         "picocolors": "^1.1.1"
       },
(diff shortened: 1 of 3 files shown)


fahertym:
## Metro family: targeted non-major update attempted, and it moves nothing

Recording the attempt rather than the inference.

```
$ npm update --package-lock-only metro metro-config metro-transform-worker image-size
$ # result:
  packages moved : 0
  vulnerabilities: 4 -> 4
```

So the four remaining `sdk/react-native` findings genuinely cannot be resolved inside the declared ranges. Why:

- `image-size` is the **sole root advisory** (GHSA-w3rx-r6r6-pgpr, GHSA-5p2g-fcmc-qvqq). Its affected range is `*` — **every published version** — so bumping `image-size` itself can never fix anything.
- `metro` is vulnerable *via* `image-size`; `metro-config` and `metro-transform-worker` are vulnerable *via* `metro`, in a mutual cycle. They are **one** finding reported four times, not four findings.
- The fix therefore has to come from Metro moving past `0.84.4`, and Metro arrives through `react-native ^0.85.1` (devDependency). The ceiling is set by React Native, not by anything this lockfile can express.

Deliberately **not** crossing the React Native major boundary to make `npm audit` print zero. That is a toolchain migration with its own verification surface, and it does not belong in a lockfile-only PR.

Lockfile restored to the state under review; this attempt left no residue.

